PR #53, opened 2014-06-25
[Alternative]: Add User Namespace support to libcontainer.

mrunalp:
Here we use container.User to get the uid/gid/suppGids in the container and then look them up in the mappings to get the corresponding uid/gid/suppGids on the host. We switch to those ids in the parent before the call to clone into new userns.

If no mapping is found, then we use the overflowuid/overflowgid.

For sample integration with docker see -- https://github.com/mrunalp/docker/compare/dev/userns_alt?expand=1
Docker-DCO-1.1-Signed-off-by: Mrunal Patel mrunalp@gmail.com (github: mrunalp)


frankscholten:
+1

PR #124, opened 2014-07-24
Initial support for multiple addresses per veth

TheDude05:
Add very basic support for having multiple IP addresses per interface. I'm looking for feedback on my approach before I dive much deeper into this as it changes the JSON format.


crosbymichael:
I"m going to close this one.  There is alot of work being done in networking for the container ecosystem and the rebase will be rough now.

Thanks for this.

PR #132, opened 2014-07-29
in order to fix the exit status handling on docker, we need to not repor...

vieux:
...t exec.ErrNotFound to the parent


crosbymichael:
This is fixed in the new API

PR #169, opened 2014-08-28
add add replace route, fix displaying ipv6 addresses and routes

vtolstov:
Sometimes we need ability to replace exiting routes with new, without deleting previous.
This patch add ability to replace route. Also has minor fixes for ipv6 addresses and ipv4/ipv6 routes.
Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>


LK4D4:
Looks like author using other netlink library now. I personally think that we should too.
Closing.

PR #187, opened 2014-09-08
nsenter can open namespaces from filesystem rather than from pid 

shykes:
This allows entering namespaces independently of the processes inside it.


crosbymichael:
This is supported in the new API for general containers.

PR #215, opened 2014-10-04
2014 10 03 dummy net strategy signed

tobert:
I threw this together for some experiments I'm running locally. Submitting this PR in case it's something you want upstream.

The use case I'm working on now is using link-local services to provide an HTTP service for container introspection similar to EC2's metadata API. I'll post separately about this service if it works out.

Another use case is adding a dummy with a fixed (link-local) address that provides host services, such as an HTTP proxy. The container can be started with a dummy interface then varnish/squid/apache/whatever can be bound to the interface using ip netns exec leaving the process invisible to the container but visible on the network with no risk of accidentally linking containers' networks.

My proposal for the metadata API: https://github.com/docker/docker/issues/8427

Edits: add link

### DIFF
--- a/network/dummy.go
+++ b/network/dummy.go
@@ -1,0 +1,62 @@
+// +build linux
+
+package network
+
+import (
+	"fmt"
+
+	"github.com/docker/libcontainer/netlink"
+)
+
+type Dummy struct {
+}
+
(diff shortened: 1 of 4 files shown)


LK4D4:
Looks like PR is abandoned. Feel free to reopen or create new PR.

PR #217, opened 2014-10-09
[RFC] Initial integration of libct into libcontainer

avagin:
Hi All,

I'm integrating libct with libcontainer. I would like to share with you
this early draft to gather your opinions and advices.

This patch adds new factory and container classes to create and manage
containers via LibCT. This version is based on Michael's version of
libcontainer, where is a new api is used.  It has minimal functionality
just to show you how it looks like.

The current version of the abstract interface is a bit different. It
doesn't affect the idea of this patch. I will update the patch as soon
as libcontainer starts using the new interface.

LibCT is a library for creating and managing containers. It has ability
to add different backends. Currently it has only a native backend, which
is the same what libcontainer has now. We are developing a backend for
OpenVZ and PCS.

Notes:
The dockerinit process isn't required for libct. If I understand
correctly, dockerinit is used, because in golang code we should make
os.exec() immediately after os.clone().

If you want to play with this code, you can clone
https://github.com/avagin/libcontainer/commits/devel-new

Cc: Michael Crosby <michael@docker.com>
Cc: Victor Marmol <vmarmol@google.com>
Cc: Pavel Emelyanov <xemul@parallels.com>

### DIFF
--- a/libct/mount.go
+++ b/libct/mount.go
@@ -1,0 +1,148 @@
+// +build linux
+
+package libct
+
+import (
+	"fmt"
+	"syscall"
+
+	_libct "github.com/xemul/libct/go"
+	"github.com/docker/libcontainer/devices"
+	"github.com/docker/libcontainer/label"
+	"github.com/docker/libcontainer/mount"
+)
+
+// default mount point flags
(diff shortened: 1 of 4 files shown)


avagin:
https://github.com/docker/libcontainer/pull/434

Issue #218, opened 2014-10-09
Separate netlink socket and message handling into separate package


hugoduncan:
In order to use the netlink support in libcontainer for netlink message types that it does not directly wrap, it would be useful to have an exported interface for creating sockets,  creating messages, parsing messages, etc.  

Is a PR to implement this likely to be considered?

LK4D4:
There is already pretty neat for of our netlink:
vishvananda/netlink

PR #220, opened 2014-10-10
Enable reuse of low-level netlink functions

hugoduncan:
Factors out basic netlink functions into the nlmsg package.

Note that this does not provide forwarding for functions previously
exported from Netlink, but now exported only from nlmsg.  While
technically a breaking change, I'm not sure if these functions could
have been meaningfully used from netlink.  I'm happy to implement
forwarding function if needed though.

Addresses #218

### DIFF
--- a/netlink/netlink_linux.go
+++ b/netlink/netlink_linux.go
@@ -1,15 +1,15 @@
 package netlink
 
 import (
-	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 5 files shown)


crosbymichael:
@hugoduncan if you or anyone else has time to rebase this then feel free to reopen this PR when it's finished. Thanks for working on this.

PR #227, opened 2014-10-17
Fix RtAttr.Len for case with children

hugoduncan:
In RtAttr.ToWireFormat, the children are written with the length aligned with
rtaAlignOf.  The length calculated in RtAttr.Len should reflect this,
otherwise the slice allocated in RtAttr.ToWireFormat is too small.


LK4D4:
I think I close this due inactivity. Feel free to reopen.

PR #263, opened 2014-11-14
Introduce support for syscall filtering in containers

rhatdan:
This PR introduces the ability to filter system calls on a per-container basis on Linux, using libseccomp to support multiple architectures.

This adds another layer of security between containers and the kernel. System calls which are unnecessary in a container or problematic from a security perspective can be restricted to prevent their use. Most of the truly problematic syscalls are already restricted by dropping capabilities; this adds an additional, finer-grained layer of protection.

There's a similar feature present in LXC already, with the significant difference that LXC uses a whitelist of system calls, whereas these patches use a blacklist. The blacklist approach ensures no difference in functionality to clients not explicitly aware of seccomp support (the restricted syscalls list in the container config is left empty, and the seccomp init function exits without taking action).

This PR adds a vendored library dependency (Go bindings for libseccomp) and a build dependency on libseccomp >= v2.1. The actual changes to libcontainer are fairly minimal, most of the delta is in the libseccomp bindings.

Presently missing: integration tests, documentation

Docker-DCO-1.1-Signed-off-by: Matt Heon <mheon@redhat.com> (github: mheon)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM golang:1.4
 
+RUN apt-get install -y libseccomp2 libseccomp-dev
 RUN go get golang.org/x/tools/cmd/cover
 
 ENV GOPATH $GOPATH:/go/src/github.com/docker/libcontainer/vendor
@@ -17,7 +18,7 @@ WORKDIR /go/src/github.com/docker/libcontainer
 RUN cp sample_configs/minimal.json /busybox/container.json
 
 RUN go get -d -v ./...
-RUN make direct-install
+RUN  TEST_TAGS="-tag seccomp" make direct-install
 
 ENTRYPOINT ["/dind"]
-CMD ["make", "direct-test"]
+CMD ["make", "TEST_TAGS=\"-tag seccomp\"", "direct-test"]
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,7 @@ all:
 	docker build -t dockercore/libcontainer .
 
 test: 
-	# we need NET_ADMIN for the netlink tests and SYS_ADMIN for mounting
-	docker run --rm -it --privileged dockercore/libcontainer
+	docker run --rm -it --privileged dockercore/libcontainer make TEST_TAGS="${TEST_TAGS}" direct-test
 
 sh:
 	docker run --rm -it --privileged -w /busybox dockercore/libcontainer nsinit exec sh
--- a/config.go
+++ b/config.go
@@ -4,6 +4,7 @@ import (
 	"github.com/docker/libcontainer/cgroups"
 	"github.com/docker/libcontainer/mount"
 	"github.com/docker/libcontainer/network"
+	"github.com/docker/libcontainer/security/seccomp"
 )
 
 type MountConfig mount.MountConfig
@@ -120,6 +121,9 @@ type Config struct {
 	// Rlimits specifies the resource limits, such as max open files, to set in the container
 	// If Rlimits are not set, the container will inherit rlimits from the parent process
 	Rlimits []Rlimit `json:"rlimits,omitempty"`
+
+	// Syscalls which will be restricted on container start
+	Seccomps []seccomp.Seccomp `json:"seccomps,omitempty"`
 }
 
 // Routes can be specified to create entries in the route table as the container is started
--- a/integration/seccomp_test.go
+++ b/integration/seccomp_test.go
@@ -1,0 +1,70 @@
+// +build seccomp,linux,cgo
+
+package integration
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/docker/libcontainer/security/seccomp"
+)
+
+func TestSeccompDenyGetcwd(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
+	rootfs, err := newRootFs()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer remove(rootfs)
+
+	config := newTemplateConfig(rootfs)
+	config.Seccomps = append(config.Seccomps, seccomp.Seccomp{Syscall: "getcwd"})
+
+	buffers, exitCode, err := runContainer(config, "", "pwd")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if exitCode != 1 {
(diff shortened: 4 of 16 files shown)


jandre:
Hi, is there any reason why this uses a blacklist-approach only? it's kind of weird.   Should the user not be able to specify whether they want to run in blacklist or whitelisting mode?

rhatdan:
jandre we could support both, but the white list would be a lot harder to put together.  
I could see where we have a drop-all and then add them back in.

jandre:
Putting a whitelist together is certainly not an easy thing (and it will vary from app to app), but I imagine you could put seccomp into permit, but log unexpected behaviors mode (similar to how you would train AppArmor or SELinux profiles by putting it in complain mode).   There's a quick demo of this here: http://outflux.net/teach-seccomp/ (see syscall-reporter).   Then, you could create tools read the logfile and build a profile automagically (e.g., something like the AppArmor easyprof tool).

Anyway, I think having it by default in blacklist mode is a good idea, but allowing the user to put it in whitelist mode gives a lot more flexibility.   Let's say you were leveraging containers as a true 'sandboxed' compute cluster, this would be a great step into making that happen securely in the future.

rhatdan:
Yes I agree, although it is difficult to figure out when you have done enough testing before putting it in non-testing mode.  SELinux and AppArmor are a little bit simpler in that I think the access is a little easier to understand.  

I don't think the patch currently allows us to specify what happens when a process is not allowed a syscall, I  think this version will return EPERM.  

@mheon do you know?

cyphar:
I have a feeling that the default set of enabled syscalls should be a whitelist, and people can add/remove from that set (just as with capabilities). We've been burned by blacklists before, let's not do it again.

pcmoore:
@jandre unfortunately the permissive/reporting mode for seccomp isn't really the same as what is exists with SELinux/AppArmor/etc.; there are a number of limitations, the most significant is that the individual container applications will likely need to be made aware of seccomp and ensure they don't overwrite the necessary signal handlers.

There is currently no *good* permissive/reporting mode for seccomp filters.

rhatdan:
@cyphar It is not that easy, and really syscall filtering is not that similar to capabilities or SELinux or Apparmor which are all about white listing.  All syscall filtering is doing is reducing the attack surface on the kernel.  There are going to be huge wholes that can not be closed, like ioctl. 

Bottom line with this technology is if we choose a black list people will use it.  We might even be able to slowly increase the blacklist.  If we choose a whitelist we will break lots of apps, and people will not take the time to figure out what to add, so they will run in --privileged mode, turning off ALL security including SELinux, Capabilites, Seccomp, UserNamespace...

The other tools I have looked at that use seccomp systemd-nspawn, qemu, and another I can not remember the name use Blacklist.

If you can really define the application that you are going to run you can really lock it down with tools like SELInux/AppArmor and Seccomp, but when you have a general purpose tool like Docker Containers, I would argue it is almost impossible to successfully define a limited whitelist.

jandre:
I 100% agree the default should be a blacklist (or maybe just have seccomp filtering disabled by default), but what I don't understand (and this is just simply a flag in libseccomp init) is why the user can't toggle between blacklist mode and whitelist mode, should they need to.   You are making it seem like we have to do one or the other.   lxc containers allow you to specify this in the first line of their config.   See:  https://github.com/lxc/lxc/blob/master/doc/examples/seccomp-v2-blacklist.conf  vs:  https://github.com/lxc/lxc/blob/master/doc/examples/seccomp-v1.conf

rhatdan:
Agreed we should make seccomp patch handle both models.  Hopefully libseccomp makes it easy to do this.

pcmoore:
@rhatdan It's easy, you just pick the default action (allow for a blacklist, kill for a whitelist) when you call seccomp_init().

rhatdan:
Ok, I will have @Mheon update the patch.

cyphar:
@rhatdan Surely if we can create a blacklist of syscalls, and we know what syscalls currently exist, then we can create a whitelist of syscalls? I understand that it might be an issue to actually create the whitelist, but when you're dealing with syscall filtering you're going to break *some* applications anyway. If you have the ability to specify which syscalls to explicitly allow (or disallow), then maintainers can trivially fix their runconfig options (it's just a matter of `strace`ing the binary). I'm just thinking forwards on this one: if (say 15 kernel revisions later) a set of new syscalls are added (and since they are new, there are probably bugs that could lead to security vulnerabilities), then those syscalls increase the attack surface by a wide margin in the interlude between the kernel releasing the new syscalls and us blacklisting them.

After all: `{whitelist} = {everything} - {blacklist}` (except the fact that a blacklist allows more things if the set of everything increases, while a whitelist blocks more things if the set of everything increases.

pcmoore:
@cyphar It is important not to underestimate the difficulty of creating, and maintaining, a syscall whitelist for an application.

cyphar:
@pcmoore For some reason, my edit to that comment isn't being shown. But yeah, I understand that maintaining a syscall whitelist would not be an easy task. As long as we support both modes, people who have the time and resources to maintain a syscall whitelist can do so.

rhatdan:
cyphar,  I would argue that your statement is the problem.  " then maintainers can trivially fix their runconfig options (it's just a matter of straceing the binary)"  is the problem. This is not the way docker works.  Docker does not support alternate runconfig per image.  It is one size fits all.  Although this is something I wish to fix in the future.  When someone write the apache image they do not know what applications the user will run on top of the image.  So how do they define a whitelist of what syscalls can or can not be run within the container.  When the end user or Admin of the container image runs the app and it fails, and it will.  They will not diagnose the problem by looking in some strange log file like /var/lib/audit/audit.log to realize the "foobar" syscall is blocked, then change their docker run command to include --security-opt seccomp:allow:foobar to get their app to run only to find out that they also needed the ABC syscall.  Lather Rince repeat, what they will
do is say docker sucks and then run docker run --privileged and call it a day.

Theoretically if we had a way to allow the image to specify the command line to be used when the container is run, and the developer of the image did enough testing then a whitelist approach might be possible,

rhatdan:
Replacing this pull request with https://github.com/docker/libcontainer/pull/384

PR #303, opened 2014-12-17
Exec process should have its own cgroup

dqminh:
### How it works,

ExecIn now requires the user to pass in a config, which contains the container config and also an optional cgroup config. If the cgroup config exists, then it will create a subcgroup under the original container's cgroup and use it to process tracking. When the main exec process exits, all other processes inside the subcgroup will be stopped/killed and the subcgroup will be removed

- [x] create sub-cgroups for each execin process
- [x] track and remove orphan process after the execin process finishes
- [x] cleanup sub-cgroups after the execin process finishes

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -2,7 +2,6 @@ package fs
 
 import (
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -93,22 +92,18 @@ func Apply(c *cgroups.Cgroup, pid int) (map[string]string, error) {
 
 // Stop stops all processes in the current cgroup
 func Stop(c *cgroups.Cgroup) error {
-	d, err := getCgroupData(c, 0)
-	if err != nil {
-		return err
-	}
-
 	// first we send SIGTERM to all tasks inside the cgroup
 	if err := Kill(c, syscall.SIGTERM); err != nil {
 		return err
 	}
 
-	// if failed to stop with only SIGTERM, we send SIGKILL
-	if err := waitStop(d, 100*time.Millisecond, 5*time.Second); err != nil {
+	// if failed to stop with only SIGTERM after 5.11s, we send SIGKILL
+	if err := waitStop(c, 10*time.Millisecond, 9); err != nil {
 		if err := Kill(c, syscall.SIGKILL); err != nil {
 			return err
 		}
-		return waitStop(d, 100*time.Millisecond, -1*time.Second)
+		// if failed to stop after 10s, timedout error is returned
+		return waitStop(c, 10*time.Millisecond, 10)
 	}
 	return nil
 }
@@ -136,46 +131,25 @@ func Kill(c *cgroups.Cgroup, signal syscall.Signal) error {
 	return nil
 }
 
-// waitStop waits for all tasks in the given cgroup tasks to stop. If they
-// failed to stop after a given timeout, then this returns an error.
-//
-// if timeout is negative then we will wait forever for the tasks to be stopped
-func waitStop(d *data, interval time.Duration, timeout time.Duration) error {
-	freezerPath, err := d.path("freezer")
-	if err != nil {
-		return err
-	}
-	tasksPath := filepath.Join(freezerPath, CgroupProcesses)
-	f, err := os.Open(tasksPath)
-	if err != nil {
-		return fmt.Errorf("failed to open %q : %s", tasksPath, err)
-	}
-	defer f.Close()
-
-	done := make(chan struct{}, 1)
-	go func() {
-		for {
-			var pid int
-			_, err = fmt.Fscan(f, &pid)
-			if err == nil || err == io.EOF {
-				done <- struct{}{}
-				return
-			}
-			time.Sleep(interval)
+// waitStop waits for all tasks in the given cgroup tasks to stop. It will check
+// a given number of times with increasing delay between tries. If the tasks
+// failed to stop then approriate error will be returned
+func waitStop(c *cgroups.Cgroup, initialWait time.Duration, tries int) error {
+	for i := 0; i < tries; i++ {
+		if i != 0 {
+			time.Sleep(initialWait)
+			initialWait *= 2
 		}
-	}()
-
-	if timeout < 0 {
-		<-done
-		return nil
+		pids, err := GetPids(c)
+		if err != nil {
+			return err
+		}
+		// all tasks in the current cgroup have been removed
+		if len(pids) == 0 {
+			return nil
+		}
 	}
-
-	select {
-	case <-done:
-		return nil
-	case <-time.After(timeout):
-		return fmt.Errorf("timedout stopping freezer.tasks %q", tasksPath)
-	}
+	return fmt.Errorf("failed to stop %q: timed out", c.Name)
 }
 
 // Symmetrical public function to update device based cgroups.  Also available
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -2,10 +2,13 @@ package fs
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
+	"syscall"
+	"time"
 
 	"github.com/docker/libcontainer/cgroups"
 )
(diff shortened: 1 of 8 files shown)


dqminh:
@crosbymichael i added the code to let the nsenter process wait until we finish putting the parent into the subcgroup. It works!

So right now i think this is completed. Please let me know if the approach is reasonable.

cc @vmarmol @vishh

dqminh:
There's also a pull request at https://github.com/docker/docker/pull/9994 with an integration test for this.

config.go
```diff
@@ -122,6 +122,24 @@ type Config struct {
 	Rlimits []Rlimit `json:"rlimits,omitempty"`
 }
 
+// ExecConfig defines configuratio options for executing a process inside a
```
vishh: s/configuratio/configuration

config.go
```diff
@@ -122,6 +122,24 @@ type Config struct {
...
+// TODO:
+// - add User to customize the uid/gid of the executing process
+type ExecConfig struct {
+	// configurations of the container
+	Container *Config `json:"container"`
+
+	// current state of the container
+	State *State `json:"state"`
+
+	// Cgroups specify specific settings for the subsystems that the execed
```
vishh: nit: s/execed/exec'ed?

cgroups/fs/apply_raw.go
```diff
@@ -88,6 +91,93 @@ func Apply(c *cgroups.Cgroup, pid int) (map[string]string, error) {
...
+	return nil
+}
+
+// Kill sends the signal to all processes in the current cgroup.
+// This assumes that the freezer cgroup is setup. It first freeze the cgroup,
+// send the signal to all processes inside the cgroup and then unfreeze it to
+// make sure that signal is sent to the processes before it has a chance to
+// do something else
+func Kill(c *cgroups.Cgroup, signal syscall.Signal) error {
+	Freeze(c, cgroups.Frozen)
```
vishh: Is it really safe to assume that freezer will be available on all systems running docker?

cgroups/fs/apply_raw.go
```diff
@@ -88,6 +91,93 @@ func Apply(c *cgroups.Cgroup, pid int) (map[string]string, error) {
...
+	if err := Kill(c, syscall.SIGTERM); err != nil {
+		return err
+	}
+
+	// if failed to stop with only SIGTERM, we send SIGKILL
+	if err := waitStop(d, 100*time.Millisecond, 5*time.Second); err != nil {
+		if err := Kill(c, syscall.SIGKILL); err != nil {
+			return err
+		}
+		return waitStop(d, 100*time.Millisecond, -1*time.Second)
```
vishh: wait forever sounds dangerous. What if the process is stuck in the kernel? How will this impact users?

namespaces/execin.go
```diff
@@ -68,8 +72,33 @@ func ExecIn(container *libcontainer.Config, state *libcontainer.State, userArgs
...
+	if config.Cgroups == nil {
+		// Enter existing cgroups of the container
+		// In this case, we cant keep track of individual execin session to
+		// clean up orphan processes, so just try to remove the main process
+		// when it fail
+		if err := EnterCgroups(state, cmd.Process.Pid); err != nil {
+			return terminate(err)
+		}
+	} else {
+		parent := container.Cgroups.Name
```
vishh: Why not `parent := filepath.Join(container.Cgroups.Parent, container.Cgroups.Name)` and get rid of the 'if' condition?

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,15 @@ void nsenter()
 			exit(1);
 		}
 	}
+	// read continue signal from fd 3. the signal is 1
+	// this is so that the parent can put the proces inside correct cgroups
+	// before it has a chance to fork
+	unsigned char s;
+	if (read(3, &s, 1) != 1 || s != '1') {
+		fprintf(stderr,
+			"nsenter: Failed to get signal from parent: %m\n");
```
vishh: Arg missing?

cgroups/fs/apply_raw.go
dqminh: Hmm, right. if the process is stuck somehow, this wont return, and the cgroup will not be cleaned up. I think that the kernel will eventually try to clean everything when the container's main process exit and the PID namespace is destroyed, but i'm not too sure.

I think probably we should just return here after sending SIGKILL without waiting ? It will probably result in an immediate failure in cleaning up cgroups after ( but we retry the removal up to 5 times so hopefully it will be cleaned up eventually ). Worst case is that you have a dangling cgroup + an uninterrupted process until the PID namespace is removed.

cgroups/fs/apply_raw.go
dqminh: Hmm, right. The original assumption in the spec is that freezer should be available ( also actually devices since `GetPids` code use it ). I think if `freezer` is not available, then we would have to:
- pick any one of the available cgroup in `subsystems` and recursively send signal to all of its processes
- and/or create our own named cgroup just for process tracking

namespaces/nsenter/nsenter.c
dqminh: iirc, `%m` is the same as `%s` with `strerror(errno)`. I'm happy to revert it back though.

crosbymichael:
@dqminh 

Were we still going to look into having the nsenter proc block on a pipe until the parent places it inside the existing cgroup?

cgroups/fs/apply_raw.go
vishh: We will have to bubble up errors to the UI somehow. I have seen cases where machines will have to be rebooted to get rid of these zombie processes. Retrying after a delay sounds like a good strategy.

cgroups/fs/apply_raw.go
vishh: I vote for having a names cgroup for the purposes of tracking. We can avoid exposing this names cgroup to containers in the future when we start supporting subcontainers.

namespaces/nsenter/nsenter.c
vishh: Ahh. This is 'C'. LGTM

crosbymichael:
Docker bug about this issue 

https://github.com/docker/docker/issues/10046

cgroups/fs/apply_raw.go
dqminh: I think a named cgroup i.e `docker` is probably ideal for process tracking, but do you think this can be deferred to a later patchset :D ? If we voted to have it, then i would prefer to have the named cgroup patch landed first.

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,15 @@ void nsenter()
 			exit(1);
 		}
 	}
+	// read continue signal from fd 3. the signal is 1
+	// this is so that the parent can put the proces inside correct cgroups
+	// before it has a chance to fork
+	unsigned char s;
+	if (read(3, &s, 1) != 1 || s != '1') {
```
dqminh: @crosbymichael i do a `read(2)` here to read 1 byte from the pipe ( which should block until it reads 1 byte iirc ), and write that byte in `execin.go`.

```
+	// finish the cgroup setup, signal to the cmd to continue
+	if _, err := parent.WriteString("1"); err != nil {
 		return terminate(err)
```

cgroups/fs/apply_raw.go
dqminh: So if i understand it correctly, the new strategy is:
- send SIGTERM to all processes
- wait for a while ( here it checks every 100 ms for up to 5s ). Do you think 5s is too small / too large ? Should it be configurable so that the process/user can specify its own waiting time ?
- if the processes do not exit after the timeout, then we send SIGKILL, and then wait for a while
- if the processes still do not exit after the timeout here, bubble up errors to user

Also `waitStop` backoff is constant here, do we need it to be exponential like in cgroup removal ?

cgroups/fs/apply_raw.go
vishh: Deferring to a later patchset sounds good. I think @crosbymichael would have more context towards expecting `freezer` cgroup to be available in all docker deployments.

cgroups/fs/apply_raw.go
vishh: Ideally, we would just try once, wait for some time and then return error if cleanup fails and let the user of libcontainer handle retries and timeouts. We currently do not have a context for exec instances which makes user retries hard. I hope the new API will provide such a context. 

For now, retrying with a backoff sounds like a good strategy. 5s sounds reasonable. We could extend it a minute too. We will have to log and/or surface enough information to let the user take appropriate actions.

dqminh:
I'm going to close this for now while working on rebasing it to the new master :heart:

PR #304, opened 2014-12-17
Adds user namespace support to libcontainer

mrunalp:
The build will fail as this depends on golang 1.4
I am sending the PR, hoping to get people to test it and give feedback.
There are still log statements for debugging that I will clean up (but keeping them now to make it easier to catch issues).

@crosbymichael @vmarmol @rjnagal PTAL and try it out :)

### DIFF
--- a/config.go
+++ b/config.go
@@ -124,6 +124,11 @@ type Config struct {
 	// AdditionalGroups specifies the gids that should be added to supplementary groups
 	// in addition to those that the user belongs to.
 	AdditionalGroups []int `json:"additional_groups,omitempty"`
+	// UidMappings is an array of User ID mappings for User Namespaces
+	UidMappings []IDMap `json:"uid_mappings,omitempty"`
+
+	// GidMappings is an array of Group ID mappings for User Namespaces
+	GidMappings []IDMap `json:"gid_mappings,omitempty"`
 }
 
 // Routes can be specified to create entries in the route table as the container is started
@@ -151,4 +156,11 @@ type Rlimit struct {
 	Type int    `json:"type,omitempty"`
 	Hard uint64 `json:"hard,omitempty"`
 	Soft uint64 `json:"soft,omitempty"`
+}
+
+// IDMap represents UID/GID Mappings for User Namespaces.
+type IDMap struct {
+	ContainerID int `json:"container_id,omitempty"`
+	HostID      int `json:"host_id,omitempty"`
+	Size        int `json:"size,omitempty"`
 }
--- a/console/console.go
+++ b/console/console.go
@@ -13,7 +13,7 @@ import (
 )
 
 // Setup initializes the proper /dev/console inside the rootfs path
-func Setup(rootfs, consolePath, mountLabel string) error {
+func Setup(rootfs, consolePath, mountLabel string, hostRootUid, hostRootGid int) error {
 	oldMask := syscall.Umask(0000)
 	defer syscall.Umask(oldMask)
 
@@ -21,7 +21,7 @@ func Setup(rootfs, consolePath, mountLabel string) error {
 		return err
 	}
 
-	if err := os.Chown(consolePath, 0, 0); err != nil {
+	if err := os.Chown(consolePath, hostRootUid, hostRootGid); err != nil {
 		return err
 	}
 
--- a/mount/init.go
+++ b/mount/init.go
@@ -25,7 +25,7 @@ type mount struct {
 
 // InitializeMountNamespace sets up the devices, mount points, and filesystems for use inside a
 // new mount namespace.
-func InitializeMountNamespace(rootfs, console string, sysReadonly bool, mountConfig *MountConfig) error {
+func InitializeMountNamespace(rootfs, console string, sysReadonly bool, hostRootUid, hostRootGid int, mountConfig *MountConfig) error {
 	var (
 		err  error
 		flag = syscall.MS_PRIVATE
@@ -58,14 +58,17 @@ func InitializeMountNamespace(rootfs, console string, sysReadonly bool, mountCon
 		return fmt.Errorf("create device nodes %s", err)
 	}
 
-	if err := SetupPtmx(rootfs, console, mountConfig.MountLabel); err != nil {
+	if err := SetupPtmx(rootfs, console, mountConfig.MountLabel, hostRootUid, hostRootGid); err != nil {
 		return err
 	}
 
 	// stdin, stdout and stderr could be pointing to /dev/null from parent namespace.
 	// Re-open them inside this namespace.
-	if err := reOpenDevNull(rootfs); err != nil {
-		return fmt.Errorf("Failed to reopen /dev/null %s", err)
+	// FIXME: Need to fix this for user namespaces.
+	if hostRootUid == 0 {
+		if err := reOpenDevNull(rootfs); err != nil {
+			return fmt.Errorf("Failed to reopen /dev/null %s", err)
+		}
 	}
 
 	if err := setupDevSymlinks(rootfs); err != nil {
--- a/namespaces/exec.go
+++ b/namespaces/exec.go
@@ -4,6 +4,7 @@ package namespaces
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
@@ -25,7 +26,7 @@ const (
 // Move this to libcontainer package.
 // Exec performs setup outside of a namespace so that a container can be
 // executed.  Exec is a high level function for working with container namespaces.
-func Exec(container *libcontainer.Config, stdin io.Reader, stdout, stderr io.Writer, console, dataPath string, args []string, createCommand CreateCommand, startCallback func()) (int, error) {
+func Exec(container *libcontainer.Config, stdin io.Reader, stdout, stderr io.Writer, console, dataPath string, args []string, createCommand CreateCommand, setupCommand SetupCommand, startCallback func()) (int, error) {
 	var err error
 
 	// create a pipe so that we can syncronize with the namespaced process and
@@ -74,14 +75,6 @@ func Exec(container *libcontainer.Config, stdin io.Reader, stdout, stderr io.Wri
 	if err := InitializeNetworking(container, command.Process.Pid, &networkState); err != nil {
 		return terminate(err)
 	}
-	// send the state to the container's init process then shutdown writes for the parent
-	if err := json.NewEncoder(parent).Encode(networkState); err != nil {
-		return terminate(err)
-	}
-	// shutdown writes for the parent side of the pipe
-	if err := syscall.Shutdown(int(parent.Fd()), syscall.SHUT_WR); err != nil {
-		return terminate(err)
-	}
 
 	state := &libcontainer.State{
 		InitPid:       command.Process.Pid,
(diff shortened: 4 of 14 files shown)


mrunalp:
@avagin Yes, we need to add support for userns to nsenter. I was thinking of doing that in follow on patches after this one gets in.

crosbymichael:
This is looking good, just installed Go 1.4 for testing

crosbymichael:
Maybe we can get this merged this week.

ping @LK4D4 @vmarmol

mrunalp:
Cool, I will rebase it and address comments.

mount/init.go
```diff
@@ -58,14 +58,17 @@ func InitializeMountNamespace(rootfs, console string, sysReadonly bool, mountCon
 		return fmt.Errorf("create device nodes %s", err)
```
avagin: @mrunalp Sorry, I don't get the idea with a setup process. Which actions can't be done from userns? Could you elaborate? Maybe you can add a comment before the SetupContainer() function.
LoadContainerEnvironment(), apparmor.ApplyProfile(), label.SetProcessLabel() affect only a current process. Now we call them for the setup process and don't call for the init process. Where are these parameters applied to the init process?
Thanks.

mrunalp: You are right about SetProcessLabel and ApplyProfile. I have to move them to init. I was aware but missed moving them back. I will post an updated patch with more comments. Thanks.

mrunalp:
Ping anyone? :)

LK4D4:
@mrunalp Hah, sorry, will review now.

console/console.go
```diff
@@ -21,7 +21,7 @@ func Setup(rootfs, consolePath, mountLabel string) error {
 		return err
 	}
 
-	if err := os.Chown(consolePath, 0, 0); err != nil {
+	if err := os.Chown(consolePath, hostRootUid, hostRootUid); err != nil {
```
LK4D4: Is this always true, that `hostRootUid == hostRootGid`?

config.go
```diff
@@ -124,6 +124,11 @@ type Config struct {
 	// AdditionalGroups specifies the gids that should be added to supplementary groups
 	// in addition to those that the user belongs to.
 	AdditionalGroups []int `json:"additional_groups,omitempty"`
+	// UidMappings is an array of User ID mappings for User Namespaces
+	UidMappings []IDMap `json:"uid_mappings,omitempty"`
```
LK4D4: According to google style guide all this things should be like `UIDMappings` and `GIDMappings`, but I'm not insist :)

namespaces/exec.go
```diff
@@ -187,6 +278,52 @@ func DefaultCreateCommand(container *libcontainer.Config, console, dataPath, ini
 	command.SysProcAttr.Pdeathsig = syscall.SIGKILL
 	command.ExtraFiles = []*os.File{pipe}
 
+	if container.Namespaces.Contains(libcontainer.NEWUSER) {
+		if container.UidMappings != nil || container.GidMappings != nil {
```
LK4D4: This checks anyway done inside function. I think we can omit them.

console/console.go
mrunalp: We could make that assumption since it doesn't make sense to have separate uid/gid mappings. (Even though we have exposed that).

config.go
mrunalp: Yeah, I went back and forth on the upstream code pr to golang but saw that when id is used in combination Uid/Gid seem to be acceptable. Also, the merged code has UidMappings / GidMappings in golang ;)

console/console.go
mrunalp: I think we can pass down hostRootgGid as well.

namespaces/exec.go
mrunalp: good catch, will fix.

namespaces/exec.go
```diff
@@ -187,6 +278,52 @@ func DefaultCreateCommand(container *libcontainer.Config, console, dataPath, ini
...
+// root: the path to the container json file and information
+// args: the arguments to pass to the container to run as the user's program
+func DefaultSetupCommand(container *libcontainer.Config, console, dataPath, init string) *exec.Cmd {
+	// get our binary name from arg0 so we can always reexec ourself
+	env := []string{
+		"console=" + console,
+		"data_path=" + dataPath,
+	}
+
+	log.Println("Console", console)
```
LK4D4: Isn't this too aggressive logging? Seems like this will be in just `stderr` of any program which uses libcontainer.

nsinit/exec.go
```diff
@@ -143,6 +144,14 @@ func startContainer(container *libcontainer.Config, dataPath string, args []stri
 		return cmd
 	}
 
+	setupCommand := func(container *libcontainer.Config, console, dataPath, init string) *exec.Cmd {
+		setupCmd = namespaces.DefaultSetupCommand(container, console, dataPath, init)
+		if logPath != "" {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("log=%s", logPath))
```
LK4D4: This all stuff is little confusing for me. Why we setup `log` for cmd in `setupCommand`. Also, I wondering why we need to do this in closure? I think we can just setup it here and pass `setupCmd` and `cmd` directly.

namespaces/exec.go
mrunalp: I plan to take out all the logging (see my first comment) once the reviews are done.

config.go
LK4D4: Hehe, okay then :)

nsinit/exec.go
mrunalp: You are right. It isn't needed. I will take this out.

namespaces/exec.go
mrunalp: I think I will take them out when I address your comments so there are fewer passes to make.

namespaces/exec.go
```diff
@@ -90,11 +94,46 @@ func Exec(container *libcontainer.Config, stdin io.Reader, stdout, stderr io.Wri
...
 	if err := libcontainer.SaveState(dataPath, state); err != nil {
 		return terminate(err)
 	}
 	defer libcontainer.DeleteState(dataPath)
 
+	// Start the setup process to setup the init process
+	if container.Namespaces.Contains(libcontainer.NEWUSER) {
+		log.Println("Starting setup")
+		setupCmd := setupCommand(container, console, dataPath, os.Args[0])
+		setupOut, _ := setupCmd.StderrPipe()
```
LK4D4: can't all this be like `out, err := setupCmd.CombinedOutput()`?

mrunalp: Yes, that should work. I went back and forth while working through it. Added to my list for today.

namespaces/exec.go
```diff
@@ -157,6 +196,58 @@ func killAllPids(container *libcontainer.Config) error {
...
+func GetHostRootUid(container *libcontainer.Config) (int, error) {
+	hostRootUid := 0
+	if container.Namespaces.Contains(libcontainer.NEWUSER) {
+		if container.UidMappings == nil {
+			return -1, fmt.Errorf("User namespaces enabled, but no user mappings found.")
+		}
+		hostRootUid, found := hostIDFromMapping(0, container.UidMappings)
+		if !found {
+			return -1, fmt.Errorf("User namespaces enabled, but no root user mapping found.")
+		} else {
```
LK4D4: `else` is not necessary, `return` too :)

namespaces/exec.go
```diff
@@ -157,6 +196,58 @@ func killAllPids(container *libcontainer.Config) error {
...
+
+	return hostRootUid, nil
+}
+
+// Converts IDMap to SysProcIDMap array and adds it to SysProcAttr.
+func AddUidGidMappings(sys *syscall.SysProcAttr, container *libcontainer.Config) {
+	if container.UidMappings != nil {
+		sys.UidMappings = make([]syscall.SysProcIDMap, len(container.UidMappings))
+		for i, um := range container.UidMappings {
+			sys.UidMappings[i].ContainerID = um.ContainerID
```
LK4D4: Oh, what a shame that we can't use `SysProcIDMap` directly

namespaces/init.go
```diff
@@ -48,6 +48,15 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 		pipe.Close()
 	}()
 
+	if container.Namespaces.Contains(libcontainer.NEWUSER) {
+		err = initUserNs(container, uncleanRootfs, consolePath, pipe, args)
```
LK4D4: If we anyway return `err`, we can just return for both branches of condition and omit `else`

namespaces/init.go
```diff
@@ -141,6 +151,85 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 	return system.Execv(args[0], args[0:], os.Environ())
 }
 
+func initUserNs(container *libcontainer.Config, uncleanRootfs, consolePath string, pipe *os.File, args []string) (err error) {
```
LK4D4: Could you describe pls what the difference between this and `initDefault`?

mrunalp: It doesn't do what is already done by setup.

mrunalp: I will add comments to both.

LK4D4: I'm asking because the look pretty similar to each other, maybe we can just wrap unneeded parts in 
```
if container.Namespaces.Contains(libcontainer.NEWUSER) {
}
```

mrunalp: Yeah, that's what I started with but right now at-least 2 such blocks are necessary and probably more in future. I thought it might be easier to read it if the two paths are separate.

LK4D4: Maybe you're right.

console/console.go
mrunalp: Fixed.

namespaces/exec.go
mrunalp: Fixed.

nsinit/exec.go
mrunalp: Fixed.

namespaces/exec.go
mrunalp: Fixed.

namespaces/exec.go
mrunalp: Fixed.

namespaces/init.go
mrunalp: Fixed.

namespaces/exec.go
mrunalp: Fixed.

mrunalp:
@LK4D4 Any more comments? :)

namespaces/init.go
```diff
@@ -141,6 +151,82 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
...
+	var networkState *network.NetworkState
+	if err := json.NewDecoder(pipe).Decode(&networkState); err != nil {
+		return err
+	}
+	// join any namespaces via a path to the namespace fd if provided
+	if err := joinExistingNamespaces(container.Namespaces); err != nil {
+		return err
+	}
+	if consolePath != "" {
+		if err := console.OpenAndDup("/dev/console"); err != nil {
```
crosbymichael: Should this be `consolePath` instead of hard coded `/dev/console`?

mrunalp: The mount setup is done before we reach here to setup /dev/console and the original consolePath isn't available anymore.

namespaces/execin.go
```diff
@@ -127,6 +129,68 @@ func FinalizeSetns(container *libcontainer.Config, args []string) error {
...
 
+// SetupContainer is run to setup mounts and networking related operations
+// for a user namespace enabled process as a user namespace root doesn't
+// have permissions to perform these operations.
+// The setup process joins all the namespaces of user namespace enabled init
+// except the user namespace, so it run as root in the root user namespace
+// to perform these operations.
+func SetupContainer(container *libcontainer.Config, args []string) error {
+	consolePath := ""
+	dataPath := args[0]
```
crosbymichael: Shouldn't this just be parameters of the function?

mrunalp: Yes, fixing.

namespaces/init.go
crosbymichael: Ic

namespaces/execin.go
mrunalp: Fixed.

crosbymichael:
LGTM

LK4D4:
LGTM

brauner:
Really looking forward to this. Out of impatience when can we expect to make use of `uid` and `gid` mappings with `docker run`?

mrunalp:
@brauner Don't know when that will land, but work has started on basic docker integration. Keep an eye on the PRs :)

brauner:
@mrunalp excellent! :)

slafs:
This is great! Can you please link a relevant PR from Docker to this one?

LK4D4:
@slafs We will link when we create it :)

TomasTomecek:
@slafs https://github.com/docker/docker/pull/12648

PR #311, opened 2014-12-23
new-api: execute a process inside an existing container 

avagin:


namespaces/nsenter/nsenter.go
```diff
@@ -5,6 +5,7 @@ package nsenter
 /*
 __attribute__((constructor)) init() {
 	nsenter();
```
dqminh: do we need to call `nsenter` here with the new implementation of `ExecIn` ? I think `nsexec` alone would suffice right, as `nsenter` will just exit right away because `argv[0]` does not start with `nsenter-` anymore.

nsinit/nsenter.go
```diff
@@ -1,84 +0,0 @@
...
-func nsenterExec(config *configs.Config, args []string) {
-	if err := namespaces.FinalizeSetns(config, args); err != nil {
-		log.Fatalf("failed to nsenter: %s", err)
-	}
-}
-
-// nsenterMknod runs mknod inside an existing container
-//
-// mknod <path> <type> <major> <minor>
-func nsenterMknod(config *configs.Config, args []string) {
```
vmarmol: We have talked about needing to add devices and things like that that may not entail running a binary inside. We don't have too many real usecases so can punt for now.

avagin:
It's called "nsenter";). Sure, we can use it. Here is a few reasons why I decided to use environment variables:
* Arguments of nsenter is a part of public api. We don't like StartInitialization(), so I decided to avoid another strange API. Environment variables help to hide this from end users.
* The code of nsexec.c is simpler than nsenter.c. We don't need to parse /proc/self/cmdline and arguments.

namespaces/nsenter/nsenter.go
avagin: @dqminh You are right. I'm going to remove nsenter in future patches.

avagin:
@vmarmol  What do you think about hiding the nsenter logic from end users? Now docker, nsinit know about nsenter and handle arguments for them.

https://github.com/docker/docker/blob/master/daemon/execdriver/native/exec.go#L29
https://github.com/docker/libcontainer/blob/master/nsinit/main.go#L36

I suggest to hide this from end users. They will need to know only about factory.StartInitialization().

vmarmol:
@avagin that sounds good. I agree we don't want users worrying about it.

vmarmol:
This LGTM, merging. Thanks @avagin!

PR #316, opened 2014-12-29
Fix vet errors

LK4D4:
I calmed `vet` about `libcontainer.Namespace` because it is very suspicious about types from other packages.


rjnagal:
LGTM, thanks for the fixes!

mrunalp:
LGTM

PR #317, opened 2015-01-05
Changes Dockerfile to use go 1.4

mrunalp:
Ran the tests locally using the image based off golang:1.4 and they succeeded.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com> (github: mrunalp)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
-FROM crosbymichael/golang
+FROM golang:1.4
 
-RUN apt-get update && apt-get install -y gcc make
 RUN go get golang.org/x/tools/cmd/cover
 
 ENV GOPATH $GOPATH:/go/src/github.com/docker/libcontainer/vendor


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM crosbymichael/golang
+FROM golang:1.4
 
 RUN apt-get update && apt-get install -y gcc make
```
tianon: These two should be in the base image already now, aren't they? :+1:

tianon: https://github.com/docker-library/golang/blob/184585232ab4d65ad4acbbd564ac648d7e6440b2/1.4/Dockerfile#L5

mrunalp: yes, you are right. Fixing.

mrunalp:
@crosbymichael @tianon PTAL. I guess we will have to merge this before drone could be updated.

tianon:
LGTM

rjnagal:
LGTM

LK4D4:
LGTM

Issue #232, opened 2014-10-22
SocketGroup is ignored if not in /etc/group even if gid specified


grossws:
I use centralized group management in system (so _nss_ resolves groups through some backend other than `files`) and docker can't find group (even if gid specified), so socket is owned by `root.root` instead of gid in `SocketGroup=<some-gid>`.

I think, docker should resolve groups via standard mechanism (not by parsing `/etc/groups`). To get group `getent group <group-name>` can be used (from glibc-common).

Maybe user management should be done using `getent passwd <user-name>`

See: https://github.com/docker/libcontainer/blob/master/user/user.go#L124.

LK4D4:
Whole idea of that package was to break dependency on libc. I think Getuser from stdlib calls that function. Probably later we will inroduce static linking, but for now we prefer to keep all programs statically linked.

grossws:
It's not a problem now since `docker -d -G gid` works fine on 1.6.

PR #319, opened 2015-01-07
Using getent to access user and group databases

dipankar:
Using getent, and then falling back to files. Should help with #232

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -2,9 +2,11 @@ package user
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 )
@@ -197,23 +199,36 @@ type ExecUser struct {
 	Home     string
 }
 
-// GetExecUserPath is a wrapper for GetExecUser. It reads data from each of the
+// GetExecUserPath is a wrapper for GetExecUser.
+// It initially attempts to get data using getent (from glibc-common).
+// If that fails then it reads data from each of the
 // given file paths and uses that data as the arguments to GetExecUser. If the
 // files cannot be opened for any reason, the error is ignored and a nil
 // io.Reader is passed instead.
 func GetExecUserPath(userSpec string, defaults *ExecUser, passwdPath, groupPath string) (*ExecUser, error) {
-	passwd, err := os.Open(passwdPath)
+
+	cmd := exec.Command("getent", "passwd")
+	stdout, err := cmd.Output()
+	passwd := bytes.NewReader(stdout)
 	if err != nil {
-		passwd = nil
-	} else {
-		defer passwd.Close()
+		passwd, err := os.Open(passwdPath)
+		if err != nil {
+			passwd = nil
+		} else {
+			defer passwd.Close()
+		}
 	}
 
-	group, err := os.Open(groupPath)
+	cmd = exec.Command("getent", "group")
+	stdout, err = cmd.Output()
+	group := bytes.NewReader(stdout)
 	if err != nil {
-		group = nil
-	} else {
-		defer group.Close()
+		group, err := os.Open(groupPath)
+		if err != nil {
+			group = nil
+		} else {
+			defer group.Close()
+		}
 	}
 
 	return GetExecUser(userSpec, defaults, passwd, group)


vmarmol:
@dipankar you'll need to sign your commits for the CI to be happy.

dipankar:
Fixed :+1:

user/user.go
```diff
@@ -197,23 +199,36 @@ type ExecUser struct {
...
-	passwd, err := os.Open(passwdPath)
-	if err != nil {
-		passwd = nil
-	} else {
-		defer passwd.Close()
+
+	cmd := exec.Command("getent", "passwd")
+	stdout, err := cmd.Output()
+	passwd := bytes.NewReader(stdout)
+	if err == nil {
```
vmarmol: != nil?

dipankar: Yea, the conditions were incorrect! Fixed and squashed the commits.

vmarmol:
Looks good overall, but @tianon and @cyphar know best :)

mrunalp:
LGTM

LK4D4:
@dipankar But this needs rebase :)

dipankar:
@LK4D4 Done!

LK4D4:
I wanted to write comment about `bytes.Buffer` as `cmd.Stdout`, but then realized this will be the same by memory consumption. Also maybe this can be extended to `key` parameter to getent later.
LGTM

tianon:
Isn't this somewhat dangerous, or at the very least unexpected for users?

The reason I ask is that the first binary named "getent" in the container's
PATH will now be invoked inside the container before what the user
specified, which IMO is surprising behavior.

So, as much as I like the idea here, I'm -1 without more convincing that
this won't be a problem or cause unexpected side effects.

cyphar:
**NOT** LGTM. This is wrong place to change the lookup code for passwd-formatted data. The correct place would be in `user/lookup_unix.go` -- and you would have to change both the `io.ReadCloser` and path interfaces (which would require temporary files, which I don't think is a good idea for a general purpose library call).

Also, I don't see what the benefit of this is. The point of `libcontainer/user` is to provide a pure Go replaceemnt for the `os/user` stdlib (which already uses the C stdlib). By executing `getent` (which is not safe, btw) you are just doing the same thing through the `getent` binary.

dqminh:
i agree with @tianon here. Depending on the external binary `getent` and its existence seems weird.

Does it make sense if we write a wrapper to use the underlying functions directly i.e `setpwent / getpwent / endpwent` etc. ?

cyphar:
@dqminh See my comment above about generic design. If you use the C functions, you can't parse container `/etc/passwd` and `/etc/group` data without being in the container. However, yours is a better solution IMO than executing a random command.

user/user.go
```diff
@@ -197,23 +199,36 @@ type ExecUser struct {
 	Home     string
 }
 
-// GetExecUserPath is a wrapper for GetExecUser. It reads data from each of the
+// GetExecUserFile is a wrapper for GetExecUser.
```
cyphar: The function name is `GetExecUserPath`?

dipankar:
@cyphar @tianon I do understand that executing getent make this sort of pointless, but as pointed out in #232 the existing solution does not get information from network sources. I was planning on initially abstracting it out to lookup_unix.go, but that would have changed quite a few things as you have pointed out. Thanks for the feedback! Will come up with a better solution.

dqminh:
@cyphar @tianon actually im not sure of why cant we use `os/user` to lookup username/id here. iirc, when user is setup, we should already be inside the container so `os/user` should lookup correctly. Then we dont have to worry about this because `os/user` does a proper lookup in the passwd database.

user/user.go
dipankar: Updated the comment

cyphar:
@dqminh The reason why the library looks the way it does is because it was designed with the goal of processing container passwd-formatted data from the host (why? initially in preperation for docker#9046, but also because it makes it a "complete" solution). Again, you are free to add some extra `GetPasswdEnt` or whatever functions that return `io.Reader`s which you can then pass to the general `io.Reader` `GetExecUser` (that is the **entire reason** that it exists, is to solve this problem).

tianon:
Also, to use "os/user" we have to link against nss (IIRC), which is why
that stdlib can't work in the general case.

PR #324, opened 2015-01-10
namespace: don't change namespaces which are not belonged to the CT 

avagin:
In a config file we specify which namespaces have to be created and which configuration have to be applied for them. Currently if a configuration is specified without a proper namespace, it is applied to the host namespace. It's unexpected behavior, isn't it?

With these patches we will get an error if a config file contains configuration without a proper namespace.

### DIFF
--- a/namespaces/init.go
+++ b/namespaces/init.go
@@ -82,11 +82,19 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 		}
 	}
 
-	if err := setupNetwork(container, networkState); err != nil {
-		return fmt.Errorf("setup networking %s", err)
-	}
-	if err := setupRoute(container); err != nil {
-		return fmt.Errorf("setup route %s", err)
+	cloneFlags := GetNamespaceFlags(container.Namespaces)
+
+	if (cloneFlags & syscall.CLONE_NEWNET) == 0 {
+		if len(container.Networks) != 0 || len(container.Routes) != 0 {
+			return fmt.Errorf("unable to apply network parameters without network namespace")
+		}
+	} else {
+		if err := setupNetwork(container, networkState); err != nil {
+			return fmt.Errorf("setup networking %s", err)
+		}
+		if err := setupRoute(container); err != nil {
+			return fmt.Errorf("setup route %s", err)
+		}
 	}
 
 	if err := setupRlimits(container); err != nil {
@@ -95,7 +103,12 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 
 	label.Init()
 
-	if err := mount.InitializeMountNamespace(rootfs,
+	// InitializeMountNamespace() can be executed only for a new mount namespace
+	if (cloneFlags & syscall.CLONE_NEWNS) == 0 {
+		if container.MountConfig != nil {
+			return fmt.Errorf("mount_config is set without mount namespace")
+		}
+	} else if err := mount.InitializeMountNamespace(rootfs,
 		consolePath,
 		container.RestrictSys,
 		(*mount.MountConfig)(container.MountConfig)); err != nil {
@@ -103,6 +116,9 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 	}
 
 	if container.Hostname != "" {
+		if (cloneFlags & syscall.CLONE_NEWUTS) == 0 {
+			return fmt.Errorf("unable to set the hostname without UTS namespace")
+		}
 		if err := syscall.Sethostname([]byte(container.Hostname)); err != nil {
 			return fmt.Errorf("unable to sethostname %q: %s", container.Hostname, err)
 		}
@@ -118,6 +134,9 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 
 	// TODO: (crosbymichael) make this configurable at the Config level
 	if container.RestrictSys {
+		if (cloneFlags & syscall.CLONE_NEWNS) == 0 {
+			return fmt.Errorf("unable to restrict access to kernel files")
+		}
 		if err := restrict.Restrict("proc/sys", "proc/sysrq-trigger", "proc/irq", "proc/bus"); err != nil {
 			return err
 		}
--- a/namespaces/utils.go
+++ b/namespaces/utils.go
@@ -36,9 +36,12 @@ func newInitPipe() (parent *os.File, child *os.File, err error) {
 }
 
 // GetNamespaceFlags parses the container's Namespaces options to set the correct
-// flags on clone, unshare, and setns
+// flags on clone, unshare. This functions returns flags only for new namespaces.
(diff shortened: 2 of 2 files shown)


namespaces/utils.go
```diff
@@ -37,8 +37,11 @@ func newInitPipe() (parent *os.File, child *os.File, err error) {
 
 // GetNamespaceFlags parses the container's Namespaces options to set the correct
 // flags on clone, unshare, and setns
-func GetNamespaceFlags(namespaces libcontainer.Namespaces) (flag int) {
+func GetNamespaceFlags(namespaces libcontainer.Namespaces, only_new bool) (flag int) {
```
LK4D4: in go usually camel case used :)

namespaces/init.go
```diff
@@ -82,6 +82,12 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 		}
 	}
 
+	cloneFlags := GetNamespaceFlags(container.Namespaces, true)
```
LK4D4: I think maybe better to do all this checks before real configuration? Maybe create simple function with all this checks.

namespaces/init.go
```diff
@@ -95,7 +101,11 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 
 	label.Init()
 
-	if err := mount.InitializeMountNamespace(rootfs,
+	if (cloneFlags & syscall.CLONE_NEWNS) == 0 {
+		if container.MountConfig != nil {
+			fmt.Errorf("mount_config is set without mount namespace")
```
LK4D4: Return miss?

xiaods:
i have refactor this pr base on @LK4D4 's suggest. have a look please.

namespaces/init.go
avagin: I was thinking about this. mount.InitializeMountNamespace() can't be executed without a mount namespace even if container.MountConfig is nil.  I would like to have these checks near the code which they affect. But I don't insist on this and if mount.InitializeMountNamespace() isn't convinced you, I can rework this series. Thanks.

avagin: @LK4D4  Another version has been suggested by @xiaods, so you can choose any of them. Thanks.

LK4D4:
@xiaods @avagin I'm not a fan of such development model. For now I suggest you to decide which PR we should leave opened, but next time I will appreciate if you wil make PRs to other's PR or just reviewing it. Thanks!

mrunalp:
Agree with @LK4D4

xiaods:
@LK4D4 got it. thanks for your comments.

avagin:
Can we continue this discuassion? @mrunalp, you probably should like this functionality, because libcontainer recently destroyed your host mount namespace;).

namespaces/utils.go
```diff
@@ -37,8 +37,11 @@ func newInitPipe() (parent *os.File, child *os.File, err error) {
 
 // GetNamespaceFlags parses the container's Namespaces options to set the correct
 // flags on clone, unshare, and setns
-func GetNamespaceFlags(namespaces libcontainer.Namespaces) (flag int) {
+func GetNamespaceFlags(namespaces libcontainer.Namespaces, onlyNew bool) (flag int) {
```
LK4D4: There is no usage for `onlyNew==false`, so I propose to remove it now and add it later if needed

avagin: Fixed, thanks.

namespaces/init.go
```diff
@@ -82,11 +82,19 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
...
 
-	if err := setupNetwork(container, networkState); err != nil {
-		return fmt.Errorf("setup networking %s", err)
-	}
-	if err := setupRoute(container); err != nil {
-		return fmt.Errorf("setup route %s", err)
+	cloneFlags := GetNamespaceFlags(container.Namespaces)
+
+	if (cloneFlags & syscall.CLONE_NEWNET) == 0 {
+		if len(container.Networks) != 0 || len(container.Routes) != 0 {
```
LK4D4: Just style nit. I don't like `else` clauses :) So I propose to combine two `if`-s and remove `else`.

namespaces/init.go
```diff
@@ -95,7 +103,11 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 
 	label.Init()
 
-	if err := mount.InitializeMountNamespace(rootfs,
+	if (cloneFlags & syscall.CLONE_NEWNS) == 0 {
```
LK4D4: Same here

mrunalp:
@avagin Ha ha, yes you make a compelling case :) Reviewing.

namespaces/init.go
mrunalp: I think that combining all the if clauses won't be that good if we add more things to what we do in a particular namespace. So, we can have two level ifs like here, but just take out the else. If we have gotten past the the ifs, then we know that it is safe to call the setup for a particular namespace.

namespaces/init.go
avagin: @LK4D4 unfortunately it is required here, because InitializeMountNamespace() has side-effects even if a config doesn't contain MountConfig. The first "else" was added just to have the same style in all places.

LK4D4: I'm okay with style, but it is not required here.
```
if (cloneFlags & syscall.CLONE_NEWNS) == 0 && container.MountConfig != nil {
    return fmt.Errorf("mount_config is set without mount namespace")
}
if err := mount.InitializeMountNamespace(rootfs,...
```
should be same.
Feel free to left it as-is, just FYI :)

mrunalp: I'm with @LK4D4, but don't mind nested ifs ;)

avagin: @LK4D4 @mrunalp I understand what you say and agree with you. I think here is another situation. @LK4D4, in your code mount.InitializeMountNamespace() is executed if cloneFlags = 0 and container.MountConfig == nil, but it must not be executed in this case. Sorry if I don't see something obvious.

LK4D4: Okay :) it deserves comment in code then
I'm also okay with else for `NEWNET` for symmetry.

LK4D4:
LGTM

namespaces/init.go
avagin: I've added a comment. Thanks.

namespaces/init.go
```diff
@@ -118,6 +134,9 @@ func Init(container *libcontainer.Config, uncleanRootfs, consolePath string, pip
 
 	// TODO: (crosbymichael) make this configurable at the Config level
 	if container.RestrictSys {
+		if (cloneFlags & syscall.CLONE_NEWNS) == 0 {
+			return fmt.Errorf("unable to restrict access to sysctl without mount namespace")
```
mrunalp: Could we change this to "unable to restrict access to kernel files"?

avagin: Fixed, thanks

mrunalp:
LGTM. (I will send a follow on PR to add this to userns code path.)

PR #325, opened 2015-01-11
namespace don't change without a proper namespace

xiaods:
avagin report excerpt:
In a config file we specify which namespaces have to be created and which
configuration have to be applied for them. Currently if a configuration is
specified without a proper namespace, it is applied to the host namespace.

this patch is refactor from avagin's pr.
https://github.com/docker/libcontainer/pull/324

Signed-off-by: Deshi Xiao <xiaods@gmail.com>

### DIFF
--- a/namespaces/exec.go
+++ b/namespaces/exec.go
@@ -138,7 +138,7 @@ func DefaultCreateCommand(container *libcontainer.Config, console, dataPath, ini
 	if command.SysProcAttr == nil {
 		command.SysProcAttr = &syscall.SysProcAttr{}
 	}
-	command.SysProcAttr.Cloneflags = uintptr(GetNamespaceFlags(container.Namespaces))
+	command.SysProcAttr.Cloneflags = uintptr(getNamespaceFlags(container.Namespaces, true))
 
 	command.SysProcAttr.Pdeathsig = syscall.SIGKILL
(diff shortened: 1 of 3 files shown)


xiaods:
ping @avagin @LK4D4

xiaods:
@avagin thanks for your review, reformat again.

xiaods:
@avagin  i have add a check for mount.InitializeMountNamespace without mount namespace, please have a check it?

avagin:
Now we can't create a container without mntns, I don't like this. Can we skip InitializeMountNamespace() if CT is created w/o mntns?

xiaods:
@avagin you concern's skip initializeMountNamespace() logic have been apply.

xiaods:
@avagin in SPEC.md line 35,
Namespaces

Flag	Enabled
CLONE_NEWPID	1
CLONE_NEWUTS	1
CLONE_NEWIPC	1
CLONE_NEWNET	1
CLONE_NEWNS	1
CLONE_NEWUSER	0

Do we need consider no mount namespace case?

xiaods:
@avagin how about think this patch? if you don't like the refactor the patch, please let me know, i will close this PR.

xiaods:
@avagin i understand your concern. let me close this pr. let me discuss with your patch.

PR #326, opened 2015-01-12
Update ROADMAP.md to correctly reflect current arch status

estesp:
Adds POWER (ppc64 and ppc64le), System z (s390x), and ARM to list of
architectures currently supported by libcontainer

Docker-DCO-1.1-Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com> (github: estesp)

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -13,4 +13,8 @@ Our goal is to make libcontainer run everywhere, but currently libcontainer requ
 
 ## Cross-architecture support
 
-Our goal is to make libcontainer run everywhere. However currently libcontainer only runs on x86_64 systems. We plan on expanding architecture support, so that libcontainer containers can be created and used on more architectures.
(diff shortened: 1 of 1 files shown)


vmarmol:
Who ensures it works on those architectures btw? I know I'm of little use there sadly :)

jfrazelle:
It would be cool if we could add all these archs to
jenkins.dockerproject.com so we could be sure nothing breaks them.....

estesp:
@jfrazelle @vmarmol I think the plan (very soon) is to have the IBM architectures as part of CI, but I'm not directly part of that work..I know we are trying to offer hardware and/or proper access to hardware to make sure it is possible.  I personally don't know much about ARM or who is enabling/testing.

jfrazelle:
cool! arm devices are pretty easy to come by I think we would just have to
set it up :D

crosbymichael:
LGTM

vmarmol:
LGTM

PR #327, opened 2015-01-12
Refactor kill all pids

crosbymichael:
This adds a sample config for running a process in the hosts PID namespaces and refactors the `killAllPids` function to avoid two calls to FindProcess


mrunalp:
LGTM

PR #328, opened 2015-01-13
Fix exit codes when dying on a signal

icecrime:
Test the process `WaitStatus` for a signal, and return an exit code of [128 + signal which killed the process](http://tldp.org/LDP/abs/html/exitcodes.html).

Fixes docker/docker#9979.

### DIFF
--- a/namespaces/exec.go
+++ b/namespaces/exec.go
@@ -17,6 +17,10 @@ import (
 	"github.com/docker/libcontainer/system"
 )
 
+const (
+	EXIT_SIGNAL_OFFSET = 128
+)
+
 // TODO(vishh): This is part of the libcontainer API and it does much more than just namespaces related work.
 // Move this to libcontainer package.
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

crosbymichael:
ping @mrunalp what do you think?

namespaces/exec.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/docker/libcontainer/system"
 )
 
+const (
+	EXIT_SIGNAL_OFFSET = 128
```
tianon: Why the C-style capitalization?

crosbymichael: Because that is how all the other consts are for signals and most of the consts in libcontainer

mrunalp:
Looking

mrunalp:
LGTM

PR #330, opened 2015-01-13
new-api: implement fs and systemd cgroup managers

avagin:


cgroups/cgroups.go
```diff
@@ -6,6 +6,19 @@ import (
 	"github.com/docker/libcontainer/devices"
 )
 
+type Manager interface {
```
vmarmol: nit: Can we add a bit of documentation to each of these?

cgroups/fs/apply_raw.go
```diff
@@ -24,6 +24,11 @@ var (
 	CgroupProcesses = "cgroup.procs"
 )
 
+type Manager struct {
+	Cgroups *cgroups.Cgroup
+	paths   map[string]string
```
vmarmol: nit: Can we add a comment for this being a map from what to what? I believe it is from cgroup subsystem to path at which it is mounted.

cgroups/fs/apply_raw.go
```diff
@@ -118,21 +141,21 @@ func GetStats(systemPaths map[string]string) (*cgroups.Stats, error) {
...
+func (m *Manager) Freeze(state cgroups.FreezerState) error {
+	d, err := getCgroupData(m.Cgroups, 0)
 	if err != nil {
 		return err
 	}
 
-	c.Freezer = state
+	m.Cgroups.Freezer = state
 
 	freezer := subsystems["freezer"]
```
vmarmol: we should check this and return an error if there is no freezer (I know it's not really related to your PR :) )

cgroups/manager/manager.go
```diff
@@ -0,0 +1,19 @@
+package manager
+
+import (
+	"github.com/docker/libcontainer/cgroups"
+	"github.com/docker/libcontainer/cgroups/fs"
+	"github.com/docker/libcontainer/cgroups/systemd"
+)
+
+func NewCgroupManager(cgroups *cgroups.Cgroup) cgroups.Manager {
```
vmarmol: Not really related to this PR, but I'm not a huge fan of how this object is not really initialized until Apply() is called. Maybe add a TODO to make this to the equivalent of Apply() at some point? Feel free to add the TODO for me.

linux_factory.go
```diff
@@ -116,7 +117,8 @@ func (l *linuxFactory) Load(id string) (Container, error) {
 		return nil, err
 	}
 
-	cgroupManager := NewCgroupManager()
+	cgroupManager := cgroups.NewCgroupManager(config.Cgroups)
+	cgroupManager.SetPaths(state.CgroupPaths)
```
vmarmol: WDYT if instead of a SetPaths we have a NewCgroupManagerWithPaths()?

vmarmol:
Minor comments, overall it looks very good to me.

cgroups/cgroups.go
mrunalp: +1

avagin:
@vmarmol I have fixed all comments. Thank you for the review.

cgroups/cgroups.go
```diff
@@ -6,6 +6,33 @@ import (
...
+	// Returns statistics for the cgroup set
+	GetStats() (*Stats, error)
+
+	// Toggles the freezer cgroup according with specified state
+	Freeze(state FreezerState) error
+
+	// Destroys the cgroup set
+	Destroy() error
+
+	// Paths maps cgroup subsystem to path at which it is mounted.
```
vmarmol: These belong in manager below?

cgroups/fs/apply_raw.go
```diff
@@ -118,21 +137,25 @@ func GetStats(systemPaths map[string]string) (*cgroups.Stats, error) {
...
-	d, err := getCgroupData(c, 0)
+func (m *Manager) Freeze(state cgroups.FreezerState) error {
+	d, err := getCgroupData(m.Cgroups, 0)
 	if err != nil {
 		return err
 	}
 
-	c.Freezer = state
-
 	freezer := subsystems["freezer"]
```
vmarmol: can we check here?

freeze, ok := subsystems["freezer"]
if !ok {
  return fmt.Errorf("the freezer subsystem is not present in this container")
}

avagin: subsystems always has the "freezer" key

```golang
        subsystems = map[string]subsystem{
                "devices":    &DevicesGroup{},
                "memory":     &MemoryGroup{},
                "cpu":        &CpuGroup{},
                "cpuset":     &CpusetGroup{},
                "cpuacct":    &CpuacctGroup{},
                "blkio":      &BlkioGroup{},
                "perf_event": &PerfEventGroup{},
                "freezer":    &FreezerGroup{},
        }
```

I think data.path() does what you suggest.
https://github.com/docker/libcontainer/blob/master/cgroups/fs/apply_raw.go#L181

vmarmol: Ah, yes you're 100% right.

cgroups/cgroups.go
avagin: NewCgroupManager() and LoadCgroupManager() require Paths and Cgroups attributes. I have updated the comment.

vmarmol:
LGTM

PR #331, opened 2015-01-13
Backup and copy the images directory over a tmpfs directory

rhatdan:
The idea here is if a user chooses to use a tmpfs over a directory, then we
want the content of the images directory to be preserved.  The first use
case I would like to use for this would be /run.  But if the --tmpfs patch or
similar got merged, I think you would need something like this patch.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/symlink"
 	"github.com/docker/libcontainer/label"
 )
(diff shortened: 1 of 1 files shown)


rhatdan:
Playing around with this, I have seen problems with potential symlinks. For example I tried this with a tmpfs on /etc, and it blew up.  We might want to change it to reporting a warning rather then an error.

mrunalp:
@rhatdan will have to vendor logrus package (from the build failures).

crosbymichael:
We cannot import the archive package in libcontainer and cannot copy files in libcontainer.

Sorry but we will not copy file like this in libcontainer.

rhatdan:
Any suggestion on how we could do this then?

PR #334, opened 2015-01-15
new-api: integration: check that a process can be executed in an existing CT

avagin:


mrunalp:
Nice test of the new API. LGTM.

PR #336, opened 2015-01-15
nsenter waits for parent signal before forking

dqminh:
This allows the parent to place the process into cgroup first before spawning the user process.

### DIFF
--- a/namespaces/nsenter/nsenter.c
+++ b/namespaces/nsenter/nsenter.c
@@ -155,6 +155,12 @@ void nsenter()
 			exit(1);
 		}
 	}
+	// blocking until the parent placed the process inside correct cgroups.
+	unsigned char s;
+	if (read(3, &s, 1) != 1 || s != '1') {
+		pr_perror("failed to receive synchronization data from parent");
+		exit(1);
+	}
 	// Setns on all supported namespaces.
 	char ns_dir[PATH_MAX];
 	memset(ns_dir, 0, PATH_MAX);
--- a/namespaces/nsenter/nsenter.c
+++ b/namespaces/nsenter/nsenter.c
@@ -28,7 +28,6 @@ void get_args(int *argc, char ***argv)
 		pr_perror("Unable to open /proc/self/cmdline");
 		exit(1);
 	}
-
 	// Read the whole commandline.
 	ssize_t contents_size = 0;
 	ssize_t contents_offset = 0;
@@ -98,13 +97,12 @@ void nsenter()
 	if (strncmp(argv[0], kNsEnter, strlen(kNsEnter)) != 0) {
 		return;
 	}
-
-	#ifdef PR_SET_CHILD_SUBREAPER
+#ifdef PR_SET_CHILD_SUBREAPER
 	if (prctl(PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0) == -1) {
 		pr_perror("Failed to set child subreaper");
 		exit(1);
 	}
-	#endif
+#endif
 
 	static const struct option longopts[] = {
 		{"nspid", required_argument, NULL, 'n'},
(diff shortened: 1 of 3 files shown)


dqminh:
cc @crosbymichael 

There's also an integration test here https://github.com/docker/docker/pull/10122

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,15 @@ void nsenter()
 			exit(1);
 		}
 	}
+	// read continue signal from fd 3. the signal is 1
+	// this is so that the parent can put the proces inside correct cgroups
+	// before it has a chance to fork
+	unsigned char s;
+	if (read(3, &s, 1) != 1 || s != '1') {
```
mrunalp: Should we check for EINTR on read here and retry?

dqminh: I think that makes sense. EINTR is the only error that we want to retry right ?

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,19 @@ void nsenter()
 			exit(1);
 		}
 	}
+	// read continue signal from fd 3. the signal is 1
+	// this is so that the parent can put the proces inside correct cgroups
```
mrunalp: typo: proces --> process.
Also, can we use something else besides the word "signal".
It is overloaded :)

dqminh: haha yes !. How about `blocking until the parent placed the process inside correct cgroups.`. I think that's sufficient.

mrunalp: Yes, that sounds better.

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,17 @@ void nsenter()
...
 	}
+	// blocking until the parent placed the process inside correct cgroups.
+	unsigned char s;
+	ssize_t ret;
+	do {
+		ret = read(3, &s, 1);
+	} while (ret == -1 && errno == EINTR);
+	if (ret != 1 || s != '1') {
+		fprintf(stderr,
+			"nsenter: Failed to get signal from parent: %m\n");
```
mrunalp: Replace "signal" by "synchronization data" here?

dqminh: fixed!

mrunalp:
Thanks for making the changes. LGTM.

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,17 @@ void nsenter()
...
 		}
 	}
+	// blocking until the parent placed the process inside correct cgroups.
+	unsigned char s;
+	ssize_t ret;
+	do {
+		ret = read(3, &s, 1);
+	} while (ret == -1 && errno == EINTR);
+	if (ret != 1 || s != '1') {
+		fprintf(stderr,
```
mrunalp: You should probably rebase and use the pr_perror macro from this commit https://github.com/docker/libcontainer/commit/35ae1c48710ff5a4db20645bc98c719cfb695b9a

namespaces/nsenter/nsenter.c
```diff
@@ -151,6 +151,17 @@ void nsenter()
 			exit(1);
 		}
 	}
+	// blocking until the parent placed the process inside correct cgroups.
+	unsigned char s;
+	ssize_t ret;
+	do {
+		ret = read(3, &s, 1);
+	} while (ret == -1 && errno == EINTR);
```
avagin: read() can return EINTR only if a process sets a signal handler without SA_RESTART. I think we can remove this loop from here. Thanks.

dqminh: @avagin Thanks for the clarification! I agree that we can remove this loop here.

mrunalp: Thanks for the pointer @avagin . I had a lingering suspicion that it might not be required in this case.

crosbymichael:
@dqminh AWESOME.  Thanks for working on this!

crosbymichael:
LGTM

crosbymichael:
LGTM

PR #338, opened 2015-01-19
cgroups: set a freezer state before calling FreezerGroup.Set()

avagin:
My previous patch moved the setting of the freezer state after the Set()
command. It's wrong, because this command uses it, so we need to set the
freezer state before the command and rollback it in an error case.

Fixes: 13a5703d853f ("cgroups: don't change a freezer state if an operation failed")


LK4D4:
LGTM

mrunalp:
LGTM

PR #339, opened 2015-01-19
new-api: implement Pause() and Resume()

avagin:


integration/exec_test.go
```diff
@@ -300,3 +303,78 @@ func TestEnter(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	container, err := factory.Create("test", config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer container.Destroy()
+
+	// Execute a first process in the container
```
mrunalp: Take out this comment as it implies there might be a second process starting after?

avagin: Fixed, thanks.

namespaces/exec.go
```diff
@@ -73,7 +71,11 @@ func Exec(args []string, env []string, console string, command *exec.Cmd, contai
 	if err != nil {
 		return terminate(err)
 	}
-	defer cgroupManager.Destroy()
+	defer func() {
+		if err != nil {
```
mrunalp: Shouldn't we always call cgroups destroy irrespective of err state on exiting this function?

avagin: No, we should not. This function returns a child pid.

mrunalp: Ahh, okay. I see what you mean (looked at the overall flow).

mrunalp:
LGTM

PR #342, opened 2015-01-21
new-api: implement Wait, WaitProcess

avagin:
and some cleanup


LK4D4:
+46  −276 looks promising :)

mrunalp:
Nice, reviewing :)

integration/exec_test.go
```diff
@@ -361,12 +380,12 @@ func TestFreeze(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if state != configs.Paused {
-		t.Fatal("Unexpected state: ", state)
-	}
 	if err := container.Resume(); err != nil {
 		t.Fatal(err)
 	}
+	if state != configs.Paused {
```
mrunalp: Why are we checking for paused after calling resume?

avagin: @mrunalp We get a state before calling resume, but check it after. I do this to not leave a container in a frozen state. Otherwise we will need to do cleanup by hand. For example here is a command from my bash_history file:
echo THAWED > /sys/fs/cgroup/freezer/system.slice/integration-test.scope/freezer.state

mrunalp: @avagin Thanks, that makes sense.

mrunalp:
LGTM

PR #345, opened 2015-01-22
Support read-only root filesystems

fabiokung:
This allows users to prepare a read only filesystem with only a few writable subpaths inside it, to be used as the root filesystem by containers.

In particular, read only loopback images can now be used as the root filesystem, allowing multiple containers to share the same base (root) image and share pages without necessarily doing CoW layers.

All changes are backwards compatible, AFAICT. More details in commit messages.

Signed-off-by: Fabio Kung <fabio@heroku.com> (github: fabiokung)

### DIFF
--- a/mount/init.go
+++ b/mount/init.go
@@ -79,7 +79,7 @@ func InitializeMountNamespace(rootfs, console string, sysReadonly bool, mountCon
 	if mountConfig.NoPivotRoot {
 		err = MsMoveRoot(rootfs)
 	} else {
-		err = PivotRoot(rootfs)
+		err = PivotRoot(rootfs, mountConfig.PivotDir)
 	}
 
 	if err != nil {
--- a/mount/mount_config.go
+++ b/mount/mount_config.go
@@ -13,6 +13,11 @@ type MountConfig struct {
 	// This is a common option when the container is running in ramdisk
 	NoPivotRoot bool `json:"no_pivot_root,omitempty"`
 
+	// PivotDir allows a custom directory inside the container's root filesystem to be used as pivot, when NoPivotRoot is not set.
+	// When a custom PivotDir not set, a temporary dir inside the root filesystem will be used. The pivot dir needs to be writeable.
+	// This is required when using read only root filesystems. In these cases, a read/writeable path can be (bind) mounted somewhere inside the root filesystem to act as pivot.
+	PivotDir string `json:"pivot_dir,omitempty"`
+
 	// ReadonlyFs will remount the container's rootfs as readonly where only externally mounted
 	// bind mounts are writtable
 	ReadonlyFs bool `json:"readonly_fs,omitempty"`
--- a/mount/pivotroot.go
+++ b/mount/pivotroot.go
@@ -10,8 +10,15 @@ import (
 	"syscall"
 )
 
-func PivotRoot(rootfs string) error {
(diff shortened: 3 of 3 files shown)


crosbymichael:
Do you think we need any type of validation around enforcing that this dir is inside the rootfs?

fabiokung:
@crosbymichael in that case MkdirAll will just fail, no?

I mean, that is a validation.

crosbymichael:
For non readonly file systems it would work, I could add `filepath.Join(rootfs, "../../../somewhere")`

fabiokung:
In that case, `pivot_root(2)` will just fail, because it requires the `pivotdir` to be inside `newroot`. From its manpages:

> new_root and put_old must not be on the same file system as the current root.

and:

> EINVAL
>   put_old is not underneath new_root.

crosbymichael:
@fabiokung thanks!  I just wanted to check to make sure.

LGTM

mrunalp:
@crosbymichael @fabiokung Yeah, I read the man page just to be sure as well. This looks safe enough. LGTM.

avagin:
Have you seen how LXC does the same without PivotDir?
https://github.com/lxc/lxc/blob/master/src/lxc/conf.c#L1027

Can we do the same here?

mrunalp:
@avagin I did in fact take a look at it while reviewing this one. Any cons against the current way of doing it? I can create a PR to do it the LXC way but not sure if it is worth it.

mrunalp:
@avagin I guess we can save creating a pivot directory for the normal flow, but would still have to do it for the readonly rootfs case.

fabiokung:
-1, without having more context on why lxc decided to do it that way.

I'm 100% sure lxc used to support a custom pivotdir in the past (some of the code and config are still even there).

AFAICT [the way it is implemented right now](https://github.com/lxc/lxc/blob/ce72f6b466/src/lxc/conf.c#L1049) doesn't respect the constraints imposed by `pivot_root(2)`, so it may be working just by accident, but I can't tell for sure without knowing more about why they did it. `man 2 pivot_root`:

> put_old **must be** underneath new_root, that is, adding a **nonzero** number of /..  to the string pointed to by put_old must yield the same directory as new_root.

mrunalp:
More details here -- https://github.com/lxc/lxc/commit/2d489f9e87fa0cccd8a1762680a43eeff2fe1b6e

fabiokung:
Thanks @mrunalp. There still isn't a lot of context on why though, and it seems to rely on internal implementation details of `pivot_root(2)` on Linux, instead of what's publicly documented. I'm still -1, but wouldn't strongly oppose if you really want to do it that way.

mrunalp:
@fabiokung Not planning to change it right now; will wait to hear from @avagin

avagin:
@fabiokung I see your point. It looks reasonable. But I don't think that the lxc solution works just by accident.

I think
>put_old must be underneath new_root, that is, adding a nonzero number of /.. to the string pointed to by put_old must yield the same directory as new_root.
should be translated as "make sure we can reach put_old from new_root"

http://permalink.gmane.org/gmane.linux.kernel.mm/7195
https://github.com/torvalds/linux/blob/master/fs/namespace.c#L2953

Thank you for the explanation.

mrunalp:
@fabiokung @avagin Thanks for a nice deep dive into pivot_root :)
Agree that docs should better reflect the behaviour for us to rely on it.

PR #346, opened 2015-01-24
Add Mode Struct to handle SELinux Relabel on Volumes

rhatdan:

### DIFF
--- a/mount/mode/mode.go
+++ b/mount/mode/mode.go
@@ -1,0 +1,80 @@
+package mode
+
+import (
+	"strings"
+)
+
+type Mode struct {
+	ModeString string `json:"modestring,omitempty"`
+}
+
+func (m *Mode) Writable() bool {
+	if strings.Index(strings.ToLower(m.ModeString), "r") != -1 && strings.Index(strings.ToLower(m.ModeString), "w") == -1 {
+		return false
+	}
+	return true
+}
+
+func (m *Mode) Relabel() bool {
+	return !(strings.Index(m.ModeString, "z") == -1 && strings.Index(m.ModeString, "Z") == -1)
+}
+
+func (m *Mode) Shared() bool {
+	return strings.Index(m.ModeString, "z") != -1
+}
+
+func (m *Mode) Private() bool {
+	return strings.Index(m.ModeString, "Z") != -1
+}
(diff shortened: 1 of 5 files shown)


rhatdan:
I have a problem with this patch, in that the mode string is not being marshalled in the json file.   I think this is causing the breakage when I try it with docker.

I can not figure out why it is not marshalling.  Any ideas?

mount/mode/mode.go
```diff
@@ -0,0 +1,80 @@
+package mode
+
+import (
+	"strings"
+)
+
+type Mode struct {
+	modeString string `json:"modeString,omitempty"`
```
mrunalp: @rhatdan Only exported fields (Public/uppercase) are marshaled by default.

rhatdan: Thank You,  Thank You , Thank You.  I have wasted more time on this.  Remind me to buy you a cookie next time I see you.  :^)

mrunalp: Sure :)

mount/mode/mode.go
```diff
@@ -0,0 +1,80 @@
+package mode
```
mrunalp: Maybe keep this a part of the mount package itself?

rhatdan: Not sure what you mean?  If you mean to not have mode be a subdir of mount, that was my original design, but I ended up with a duplicate import in label.  It can't import mount, since mount imports label.

rhatdan:
@crosbymichael @mrunalp Can someone look at this?

mount/mode/mode.go
mrunalp: I meant keep it in the mount package itself.

rhatdan: You can't do this, because we need access to the Mount.Mode within the label.go which is included by mount, so we end up with a circular dependency.

mrunalp:
Refactoring is going on in the api branch (which is close to being done). You might want to take a look at that branch. See #375

rhatdan:
I see this is merged but I am not seeing the changes in my pool.

rhatdan:
Looks like this is rejected for the original solution, which is fine with me.

PR #349, opened 2015-01-26
Use Wait4 instead of cmd.Wait

LK4D4:
Issue with cmd.Wait is that it is waiting for closing pipes and if we
have forked processes which inherited pipes from parent, then we need to
kill them to unblock cmd.Wait.

Should fix docker/docker#10303

Now idea is next:
- Wait4 for init process dead
- Kill remaining processes in cgroup (pipes closed as side effect)
- use cmd.Wait for waiting pipes flushed to client

Signed-off-by: Alexander Morozov <lk4d4@docker.com>

### DIFF
--- a/namespaces/exec.go
+++ b/namespaces/exec.go
@@ -51,10 +51,20 @@ func Exec(container *libcontainer.Config, stdin io.Reader, stdout, stderr io.Wri
 	}
 	child.Close()
 
+	wait := func() (*os.ProcessState, error) {
+		ps, err := command.Process.Wait()
+		// we should kill all processes in cgroup when init is died if we use
+		// host PID namespace
+		if !container.Namespaces.Contains(libcontainer.NEWPID) {
+			killAllPids(container)
+		}
+		return ps, err
+	}
+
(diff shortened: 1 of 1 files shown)


namespaces/exec.go
```diff
@@ -122,16 +133,14 @@ func Exec(container *libcontainer.Config, stdin io.Reader, stdout, stderr io.Wri
...
-		killAllPids(container)
+	waitStatus, err := wait()
+	if err != nil {
+		return -1, err
 	}
+	// waiting for pipe flushing
+	command.Wait()
 
-	waitStatus := command.ProcessState.Sys().(syscall.WaitStatus)
+	//waitStatus := ps.Sys().(syscall.WaitStatus)
```
crosbymichael: delete?

avagin:
Can we use cmd.Process.Wait in this case?

LK4D4:
@avagin Yeah, it is probably better to use more high-level API if we can.
I changed patch.

avagin:
LGTM

crosbymichael:
LGTM

PR #350, opened 2015-01-27
Add enviornment variable `DISABLE_SYSTEMD_CGROUPS` so you can optionally...

jfrazelle:
... not use systemd cgroups.

Just because a user uses systemd doesn't mean they want to use systemd's
cgroups.

### DIFF
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -41,6 +41,10 @@ func newProp(name string, units interface{}) systemd.Property {
 }
 
 func UseSystemd() bool {
+	if disableSystemd := os.Getenv("DISABLE_SYSTEMD_CGROUPS"); disableSystemd != "" {
+		return false
+
+	}
 	s, err := os.Stat("/run/systemd/system")
 	if err != nil || !s.IsDir() {
 		return false


jfrazelle:
This is kinda a RFC, except I really want this and I have a feeling @ibuildthecloud wants it too

mrunalp:
This sounds good to me but eventually systemd distributions won't be able to use cgroups directly (need to confirm). @vmarmol @rjnagal ping.

jfrazelle:
sorry the drone failure was actually my fault but in a different way (trying out a new image on the server) I fixed it

rjnagal:
It will change eventually, but that still seems pretty far. And systemd has
lot of pieces missing to be useful as a cgroups API right now. Michael had
proposed to go the other way and make filesystem cgroups default. But this
approach is better as it won't break all the deployments that are expecting
containers to show under the systemd specific path.

I am fine with this for now.

LK4D4:
I don't know, seems like libcontainer should work on `per-container` level. Why not add field to config? Env looks very "workaroundish" for now and I'm not sure that configuring libcontainer by env is good idea.

jfrazelle:
I can make it in the config, I am just glad that people are in favor of
this idea! Then maybe a daemon flag in docker? I NEED THIS hehe :D

On Tue, Jan 27, 2015 at 4:10 PM, Alexander Morozov <notifications@github.com

LK4D4:
@jfrazelle Who then will test systemd? :cat:
Here we discussing change to `libcontainer` and I think it is probably should be non-global flag, but flag in `libcontainer.Config`.
ping @crosbymichael

crosbymichael:
I would rather have a field in the Cgroup struct that takes the implementation to use for this.  Then the change to set this value can happen in the calling code and not have a magic switch in libcontainer.

jfrazelle:
coool coool i will modify!

LK4D4:
@crosbymichael good for me, too.

rjnagal:
+1. I like the ability to not involve systemd. Any means to do that is fine.

@vmarmol this will likely break cAdvisor.

On Tue, Jan 27, 2015 at 4:39 PM, Alexander Morozov <notifications@github.com

vishh:
+1

jfrazelle:
I'll close and reopen when I have my new  implementation :)

ibuildthecloud:
The environment variable approach does that the nice advantage that one could use it without having to go through the hassle of approval in Docker.  While everyone here agrees, I don't know if the discussion in Docker will be as amicable.  But for the docker discussion, I think a daemon flag for -cgroup-driver auto|systemd|cgroupfs would make sense.

tianon:
Sounds like it might be time for `--exec-opt` to mirror `--storage-opt` in Docker.

jfrazelle:
Ya Darren I made that exact flag :) great minds think alike

vmarmol:
Y U wanna break cAdvisor :(

More seriously, +1 on this as well (late to the party). Just to bikeshed, should this be a init-time configuration? As in, when we do libcontainer.New() we can tell libcontainer that all the containers it creates are cgroupfs or systemd. Seems like that is more what we'd like to do. On the other hand, having it on the config makes it easier for an outside party (like cAdvisor) to know what is being used.

ibuildthecloud:
I don't think it should be at init time because this is a per-container level configuration.  Obviously though from the Docker approach we will make global, but libcontainer is about having all the cool knobs.

crosbymichael:
+1 to what @vmarmol it should be at init time and not per container because that introduces major inconsistencies for any systems managing containers and reporting of a host.  There would be no reason to run systemd and cgfs on the same machine.

LK4D4:
Hmm, I don't see what can be bad with this inconsistencies for separate containers - they shouldn't intersect. Also I don't like global configurations for libs - high level object is container, not god-object `libcontainer`. Just my opinion.

ibuildthecloud:
@crosbymichael good point.  What if it was at init time but then which cgroup impl was used was stored in the state.json.  Best of both worlds?

LK4D4:
We have some `Factory` type, which is not used. Maybe should be there?

crosbymichael:
@LK4D4 yes, that is basically what @vmarmol was proposing

ibuildthecloud:
Looking at code I have, I always look for stats per container and then for each container I basically hard code to look in system.slice/docker-%s.scope and then look in docker/%s.  I don't see much problem if systemd and cgroupfs were mixed on the same system as long as I know per container which style is used.  Anyways, I don't care what color the bike shed is.

crosbymichael:
FYI for people joining this issue, it is not about supporting systemd or not supporting it.  It's about the technical facts that their cgroup implementation is not complete and ready for container workloads and we have to jump through hoops to integrate the two with half of the necessary cgroups being setup by systemd they the other half still being set manually via the filesystem code.  

The systemd effort is taking alot of the future efforts and changes into account and being cautious about that they introduce in the API but the realities and needs today for running containers are not able to be accomplished fully right now.

LK4D4:
@crosbymichael Yeah, just thought that it will be like `libcontainer.Systemd = false` cause factory is not used :)

jfrazelle:
Hmm well maybe we should decide on per container or on init, I don't mind either way :)

mrunalp:
+1 to either way. Just got pinged about some setting not available in systemd cgroups.

PR #352, opened 2015-01-28
Allow a container's cgroups driver to be configured.

jfrazelle:
This adds a `Driver` to the `Cgroups` struct. I am unsure if this is the preferred way to implement this. There is a previous discussion on https://github.com/docker/libcontainer/pull/350 as to whether this should be global or per container. I am open to other implementations :D 
Also I tried to have a `func (c *Cgroup) GetDriver()` function to cleanup all the `if container.Cgroups.Driver == "systemd" && systemd.UseSystemd()` but because it required calling `systemd.UseSystemd()` in cgroups.cgroups.go, there was a circular dependency problem :/
So open to other ideas.

### DIFF
--- a/cgroups/cgroups.go
+++ b/cgroups/cgroups.go
@@ -54,4 +54,5 @@ type Cgroup struct {
 	BlkioWeight       int64             `json:"blkio_weight,omitempty"`       // Specifies per cgroup weight, range is from 10 to 1000.
 	Freezer           FreezerState      `json:"freezer,omitempty"`            // set the freeze value for the process
 	Slice             string            `json:"slice,omitempty"`              // Parent slice to use for systemd
+	Driver            string            `json:"driver,omitempty"`             // Configure the driver that should manage the container's cgroups (default: cgroupfs, systemd)
 }
--- a/namespaces/exec.go
+++ b/namespaces/exec.go
@@ -156,7 +156,7 @@ func killAllPids(container *libcontainer.Config) error {
 		freeze  = fs.Freeze
 		getPids = fs.GetPids
 	)
-	if systemd.UseSystemd() {
+	if container.Cgroups.Driver == "systemd" {
 		freeze = systemd.Freeze
 		getPids = systemd.GetPids
 	}
@@ -327,7 +327,7 @@ func DefaultSetupCommand(container *libcontainer.Config, console, dataPath, init
 func SetupCgroups(container *libcontainer.Config, nspid int) (map[string]string, error) {
 	if container.Cgroups != nil {
 		c := container.Cgroups
-		if systemd.UseSystemd() {
+		if container.Cgroups.Driver == "systemd" {
 			return systemd.Apply(c, nspid)
 		}
 		return fs.Apply(c, nspid)
--- a/nsinit/pause.go
+++ b/nsinit/pause.go
@@ -39,7 +39,7 @@ func toggle(state cgroups.FreezerState) error {
 		return err
 	}
 
-	if systemd.UseSystemd() {
+	if container.Cgroups.Driver == "systemd" {
 		err = systemd.Freeze(container.Cgroups, state)
 	} else {
 		err = fs.Freeze(container.Cgroups, state)


namespaces/exec.go
```diff
@@ -156,7 +156,7 @@ func killAllPids(container *libcontainer.Config) error {
 		freeze  = fs.Freeze
 		getPids = fs.GetPids
 	)
-	if systemd.UseSystemd() {
+	if container.Cgroups.Driver == "systemd" && systemd.UseSystemd() {
```
crosbymichael: I think we should remove the `systemd.UseSystemd()` call here.  This can be called by the caller, i.e. Docker, to determine what value to set as the Cgroups.Driver.  

If you set the driver to systemd and you don't have systemd enabled then it should fail hard because it's the callers fault and we cannot honor what they requested.

jfrazelle: perf will do

cgroups/cgroups.go
```diff
@@ -54,4 +54,5 @@ type Cgroup struct {
 	BlkioWeight       int64             `json:"blkio_weight,omitempty"`       // Specifies per cgroup weight, range is from 10 to 1000.
 	Freezer           FreezerState      `json:"freezer,omitempty"`            // set the freeze value for the process
 	Slice             string            `json:"slice,omitempty"`              // Parent slice to use for systemd
+	Driver            string            `json:"driver,omitempty"`             // Configure the driver that should manage the container's cgroups (cgroupfs or systemd)
```
rjnagal: Mention that cgroupfs is the default.

vmarmol:
LGTM

cgroups/cgroups.go
jfrazelle: sorry about the slow update, updated now :)

calavera:
Reading through the `UseSystemd()` function, it looks like it doesn't really use anything imported from the `cgroups` package https://github.com/jfrazelle/libcontainer/blob/configure-cgroup-manager/cgroups/systemd/apply_systemd.go#L43-L72, but I might be very very wrong.

Have you considered moving that function somewhere else?

jfrazelle:
I have tried :( and failed about ten times

calavera:
I guess it would involve a bigger change since there is that global connection there :/

jfrazelle:
systemd.New() is the prob iirc
https://github.com/jfrazelle/libcontainer/blob/configure-cgroup-manager/cgroups/systemd/apply_systemd.go#L54

calavera:
sometimes good is better than perfect and this PR does exactly what's intended to do :sparkles:  :+1: :sparkles:

jfrazelle:
ping @crosbymichael save me from the systemd cgroups ;)

namespaces/exec.go
```diff
@@ -327,7 +327,7 @@ func DefaultSetupCommand(container *libcontainer.Config, console, dataPath, init
 func SetupCgroups(container *libcontainer.Config, nspid int) (map[string]string, error) {
 	if container.Cgroups != nil {
 		c := container.Cgroups
-		if systemd.UseSystemd() {
+		if container.Cgroups.Driver == "systemd" && systemd.UseSystemd() {
```
crosbymichael: We should remove the ` && systemd.UseSystemd()` from this and the one below like you did on the one above

jfrazelle: updated :D

dqminh:
LGTM

ibuildthecloud:
@jfrazelle What happened to this?

jfrazelle:
I will reopen! Sorry the bday stuff was killin me before

crosbymichael:
This is implemented at the factory level and just needs a docker daemon flag to set it

jfrazelle:
Hell ya!

thaJeztah:
>This is implemented at the factory level and just needs a docker daemon flag to set it

@jfrazelle is there an issue already in docker to implement the flag? (Just to it's not forgotten)

jfrazelle:
I was going to make a patch on monday

jfrazelle:
it's not really worth of an issue needs a whole proposal ;)

thaJeztah:
:+1: good, just read this one and saw that no (open) issue was linked from the docker repo.

Happy Easter btw! Enjoy your weekend :chicken: :egg: :rabbit:

PR #357, opened 2015-02-02
Flatten config structures and remove namespace package

crosbymichael:


container.go
```diff
@@ -80,25 +89,24 @@ type Container interface {
...
-	Wait() (exitStatus int, err error)
+	Wait() (exitStatus syscall.WaitStatus, err error)
 
-	// WaitProcess waits on a process owned by the container.
+	// OOM returns a read-only channel signaling when the container receives an OOM notification.
 	//
 	// errors:
-	// ContainerDestroyed - Container no longer exists,
 	// Systemerror - System error.
-	WaitProcess(pid int) (exitStatus int, err error)
```
mrunalp: Is there a new equivalent for this?

configs/state.go
```diff
@@ -17,21 +9,30 @@ type State struct {
...
-const (
-	// The name of the runtime state file
-	stateFile = "state.json"
+// The status of a container.
+type Status int
 
+const (
 	// The container exists and is running.
-	Running RunState = iota + 1
+	Running Status = iota + 1
```
mrunalp: +1 to this rename.

container.go
crosbymichael: os.FindProcess(); proc.Wait()

linux_container.go
```diff
@@ -121,76 +315,275 @@ func (c *linuxContainer) updateStateFile() error {
...
+	panic("unreachable")
 }
 
-func (c *linuxContainer) Signal(pid, signal int) error {
-	glog.Infof("sending signal %d to pid %d", signal, pid)
-	panic("not implemented")
-}
+// finalize expects that the setns calls have been setup and that is has joined an
+// existing namespace
+func finalizeSetns(container *configs.Config) error {
```
mrunalp: Maybe rename this to finalizeContainerSetup?

linux_container.go
```diff
@@ -121,76 +315,275 @@ func (c *linuxContainer) updateStateFile() error {
...
+	// clear the current processes env and replace it with the environment defined on the container
+	if err := loadContainerEnvironment(container); err != nil {
+		return err
+	}
+
+	if err := setupRlimits(container); err != nil {
+		return fmt.Errorf("setup rlimits %s", err)
+	}
+
+	if err := finalizeNamespace(container); err != nil {
```
mrunalp: Not been a fan of this name either but can't think of a good one.

container.go
mrunalp: cool!

container.go
```diff
@@ -16,15 +26,14 @@ type Container interface {
 	// Returns the ID of the container
 	ID() string
 
-	// Returns the current run state of the container.
+	// Returns the current statusof the container.
```
dqminh: nit: status of

linux_factory.go
```diff
@@ -163,16 +214,346 @@ func (l *linuxFactory) loadContainerState(root string) (*configs.State, error) {
...
+	config := process.Config
+	networkState := process.NetworkState
+
+	// TODO: move to validation
+	/*
+		rootfs, err := utils.ResolveRootfs(uncleanRootfs)
+		if err != nil {
+			return err
+		}
+	*/
```
dqminh: should we resolve this TODO in this PR ? Where do you think the validation should be done ? in `StartInitialization` ?

linux_container.go
```diff
@@ -83,36 +96,217 @@ func (c *linuxContainer) Stats() (*ContainerStats, error) {
...
+
+func (c *linuxContainer) Signal(signal os.Signal) error {
+	glog.Infof("sending signal %d to pid %d", signal, c.state.InitPid)
+	panic("not implemented")
+}
+
+func (c *linuxContainer) Wait() (syscall.WaitStatus, error) {
+	var status syscall.WaitStatus
+	// TODO : close exec.Cmd pipes, fix in master
+	_, err := syscall.Wait4(c.state.InitPid, &status, 0, nil)
```
avagin: I don't like Wait4 here for a few reasons.
* it can be used only once, then it will return an error.
* we provide pid to an user, so he can call wait4 too.
* it doesn't work for containers without pidns.
We can try to use a cgroup release agent here...
I don't suggest to fix this issue right now, It's just for information. Maybe we need to add todo here.

process.go
```diff
@@ -21,7 +17,4 @@ type ProcessConfig struct {
 	Stdin  io.Reader
 	Stdout io.Writer
 	Stderr io.Writer
-
-	// Console is the path to the pty slave for use by the master
-	Console string
```
avagin: Could you explain why Console and Env should be only in a container config? I have only one explanation that nobody asked about this functionality.  If it's the reason, it's ok.

mrunalp: I think this is a valid point. Also, we could have more process specific parameters like capabilities, user, rlimits, etc. Should we have duplicate fields of what is in container config here and these take precedence? Or go back to not flattening it so it can be completely overriden per process?

crosbymichael: The general idea is that the container configuration specifies **how** a container and it's processes execute.  The console, env, capabilities, rlimits, etcd.  Then the process specifies **what** is run.  Args.  I guess we can make the argument that env should be removed from the container and put at the process level as well as user( maybe ).

linux_factory.go
crosbymichael: I was thinking we can do it in another pr at the `factory.Create()` time before anything else is done.

linux_container.go
crosbymichael: So do you think we should not have a wait on the container at all?

linux_container.go
crosbymichael: Ya, i was going to take another pass after this one is merged in refineing more func and method names to make sense.

mrunalp:
Merging this for now. We can work out the details as keep iterating on the new api. LGTM.

process.go
avagin: @crosbymichael  @mrunalp  From my point of view, a process config should contain (env, uid, gid, caps, tty, lsm or selinux labels, etc). A container config should contain configuration fror namespaces and cgroups and default parameters for processes.
Now we have a container config with default parameters. In future we will be able to expand the process config, when it's required.

crosbymichael: @avagin that sound good to me

mrunalp: :+1:

PR #358, opened 2015-02-03
namespaces: allow to use pid namespace without mount namespace

avagin:
The gocapability package uses /proc/PID/status to get a bounding set.
If a container uses pidns without mntns, it sees /proc from the host
namespace, but the process doesn't know its own pid in this namespace.

In this case it can use /proc/self/status, which is always the right one.

### DIFF
--- a/security/capabilities/capabilities.go
+++ b/security/capabilities/capabilities.go
@@ -1,8 +1,6 @@
 package capabilities
 
 import (
-	"os"
-
 	"github.com/syndtr/gocapability/capability"
 )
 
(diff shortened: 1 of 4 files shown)


security/capabilities/capabilities.go
```diff
@@ -11,7 +9,7 @@ const allCapabilityTypes = capability.CAPS | capability.BOUNDS
 // DropBoundingSet drops the capability bounding set to those specified in the
 // container configuration.
 func DropBoundingSet(capabilities []string) error {
-	c, err := capability.NewPid(os.Getpid())
+	c, err := capability.NewPid(0)
```
crosbymichael: what does 0 mean here?

avagin: Change capabilities of the calling process. os.Getpid() and 0 are the same except the case, when you try to access /proc from another pidns

LK4D4: Seems like drone failing because of this

avagin: @LK4D4 I've commited changes in the vendor direcotry

mrunalp:
LGTM

vmarmol:
LGTM

PR #359, opened 2015-02-04
cgroups: systemd: set DefaultDependencies=false if possible

philips:
The root problem this fixes is the docker daemon uses DefaulDependencies
for all of its scopes which means that the containers get killed by
systemd before the docker daemon is notified to shutdown. This means
that a docker run in a service file won't get ordered properly on
shutdown! This has affected many CoreOS users and is documented in
systemd as so:

"Unless DefaultDependencies=false is used, scope units will implicitly
have dependencies of type Conflicts= and Before= on shutdown.target."

Unfortunately, systemd didn't allow setting DefaultDependencies=false on
transient units until today:

    systemd-run --scope --property="DefaultDependencies=false" /usr/bin/sleep 50000
    Unknown assignment DefaultDependencies=false.
    Failed to create message: Invalid argument

Fixed here:
http://cgit.freedesktop.org/systemd/systemd/commit/?id=261420ba2a20305ad271b6f5f380aa74c5c9dd50

Discussion with systemd upstream:
http://lists.freedesktop.org/archives/systemd-devel/2014-December/026313.html
http://lists.freedesktop.org/archives/systemd-devel/2015-February/027890.html

Tested with docker and systemd master as of today and it work for me.

Signed-off-by: Brandon Philips <brandon.philips@coreos.com>

### DIFF
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -28,9 +28,10 @@ type subsystem interface {
 }
 
 var (
-	connLock              sync.Mutex
-	theConn               *systemd.Conn
-	hasStartTransientUnit bool
+	connLock                        sync.Mutex
+	theConn                         *systemd.Conn
+	hasStartTransientUnit           bool
+	hasTransientDefaultDependencies bool
 )
 
 func newProp(name string, units interface{}) systemd.Property {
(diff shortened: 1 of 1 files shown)


cgroups/systemd/apply_systemd.go
```diff
@@ -67,6 +68,17 @@ func UseSystemd() bool {
 				}
 			}
 		}
+
+		// Assume StartTransientUnit on a scope allows DefaultDependencies
```
jonboulle: Cleaner to short circuit before here if hasStartTransientUnit is false.

philips: Good point, fixed.

jfrazelle: Thanks for tracking this down Brandon, seems like a crazy one.

crosbymichael:
One question, does this depend on a specific version of systemd?

philips:
@crosbymichael It shouldn't depend on a particular version of systemd. The code in there should fall back to the old behavior of not setting DefaultDependencies if the version of systemd that is running doesn't support that.

philips:
@crosbymichael I can make this patch much shorter by making libcontainer depend on systemd master :-P

erikh:
heh

LK4D4:
LGTM

mrunalp:
I will test this out on Fedora.

mrunalp:
LGTM

PR #360, opened 2015-02-04
integration: check a container with userns

avagin:

### DIFF
--- a/integration/exec_test.go
+++ b/integration/exec_test.go
@@ -12,6 +12,17 @@ import (
 )
 
 func TestExecPS(t *testing.T) {
+	testExecPS(t, false)
+}
+
+func TestUsernsExecPS(t *testing.T) {
+	if _, err := os.Stat("/proc/self/ns/user"); os.IsNotExist(err) {
+		t.Skip("userns is unsupported")
+	}
+	testExecPS(t, true)
+}
+
+func testExecPS(t *testing.T, userns bool) {
 	if testing.Short() {
 		return
 	}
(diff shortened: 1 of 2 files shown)


integration/exec_test.go
```diff
@@ -12,6 +12,17 @@ import (
 )
 
 func TestExecPS(t *testing.T) {
+	testExecPS(t, false)
+}
+
+func TestUsernsExecPS(t *testing.T) {
+	if _, err := os.Stat("/proc/self/ns/user"); os.IsNotExist(err) {
```
LK4D4: Hmm, I don't see that we using such skip strategy for other namespaces, maybe we should just add test mappings and NEWUSER to `newTemplateConfig`?

mrunalp: I think this check is for whether the kernel supports it or not. @avagin could you confirm?

LK4D4: Yeah, I know, but other namespaces can be disabled in kernel too. Why `NEWUSER` is special? I think it is probably better to say that if you want to run tests - you should use proper kernel config, because I think `t.Skip`s can obscure some errors from developer and we will have to deal with them in our CI.

mrunalp: @LK4D4 Most likely because we assume that the other namespaces are available whereas NEWUSER was only enabled over the last year in most distributions. RHEL 7 did not have it enabled for e.g. but 7.1 will. I agree that testing should be better integrated with the kernel config.

avagin: @LK4D4 @mrunalp  userns is a special one, because it was added recently. Many distributives provide kernels where it's disabled.
Actually many people uses containers without userns. And when we enable userns, the containers is created by another way, so we need to test both configurations.

mrunalp: I think this is fine for the short term. Another option might be to have two tests; one for userns and one without. @LK4D4 WDYT?

LK4D4:
@avagin Sorry for long response. LGTM

PR #361, opened 2015-02-04
fix typo for GetHostRootGid

hqhq:

### DIFF
--- a/namespaces/exec.go
+++ b/namespaces/exec.go
@@ -192,7 +192,7 @@ func hostIDFromMapping(containerID int, uMap []libcontainer.IDMap) (int, bool) {
 	return -1, false
 }
 
-// Gets the root uid for the process on host which could be non-zero
+// Gets the root gid for the process on host which could be non-zero
 // when user namespaces are enabled.
 func GetHostRootGid(container *libcontainer.Config) (int, error) {
 	if container.Namespaces.Contains(libcontainer.NEWUSER) {
@@ -206,7 +206,7 @@ func GetHostRootGid(container *libcontainer.Config) (int, error) {
 		return hostRootGid, nil
 	}
 
-	// Return default root uid 0
+	// Return default root gid 0
 	return 0, nil
 }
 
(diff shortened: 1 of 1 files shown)


namespaces/exec.go
```diff
@@ -255,7 +255,7 @@ func AddUidGidMappings(sys *syscall.SysProcAttr, container *libcontainer.Config)
 //
 // console: the /dev/console to setup inside the container
 // init: the program executed inside the namespaces
-// root: the path to the container json file and information
+// dataPath: the path saving container's date
```
mrunalp: It isn't for date. It is for state.

mrunalp: I think it is a typo you meant data :)

mrunalp:
Also, maybe squash the commits.

hqhq:
Updated.

namespaces/exec.go
```diff
@@ -255,7 +255,7 @@ func AddUidGidMappings(sys *syscall.SysProcAttr, container *libcontainer.Config)
 //
 // console: the /dev/console to setup inside the container
 // init: the program executed inside the namespaces
-// root: the path to the container json file and information
+// dataPath: the path saving container's states
```
mrunalp: Could you change it to something like
dataPath: the path to the directory under which the container's state file is stored

namespaces/exec.go
```diff
@@ -294,11 +294,9 @@ func DefaultCreateCommand(container *libcontainer.Config, console, dataPath, ini
 // DefaultSetupCommand will return an exec.Cmd that joins the init process to set it up.
 //
 // console: the /dev/console to setup inside the container
+// dataPath: the path saving container's states
```
mrunalp: Same here

hqhq: Updated.

mrunalp:
@hqhq Thanks! LGTM

rjnagal:
LGTM.

Thanks for the cleanup.

PR #362, opened 2015-02-06
Retry getting the cgroup root at apply time.

vmarmol:
This will allow late-binding of the cgroup hierarchy.

Signed-off-by: Victor Marmol <vmarmol@google.com>

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -1,11 +1,11 @@
 package fs
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
+	"sync"
 
 	"github.com/docker/libcontainer/cgroups"
 )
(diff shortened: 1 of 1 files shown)


vmarmol:
@jaybuff

LK4D4:
Maybe we can then remove `init` and just compute root lazyli

vmarmol:
Good idea, done.

jaybuff:
perhaps include in the commit message that this resolves https://github.com/docker/docker/issues/8791

vmarmol:
@jaybuff added it to the commit.

cgroups/fs/apply_raw.go
```diff
@@ -27,18 +27,20 @@ var (
...
 	cpuRoot, err := cgroups.FindCgroupMountpoint("cpu")
 	if err != nil {
-		return
+		return err
 	}
 	cgroupRoot = filepath.Dir(cpuRoot)
 
 	if _, err := os.Stat(cgroupRoot); err != nil {
-		return
+		return err
```
LK4D4: Don't know how it was before, but if `cgroupRoot` is not `""` and `os.Stat` failed, then second time `setCgroupRoot` won't be called. Also this function now can be called from concurrent goroutines.
So, resume:
* we need set cgroupRoot back to `""` on `Stat` fail
* we need to protect it with mutex

vmarmol: Synchronized access and generation of cgroupRoot.

LK4D4:
LGTM

mrunalp:
LGTM

PR #363, opened 2015-02-07
add MemoryLimit to MemoryStats

hqhq:


vmarmol:
This is part of the container configuration, not part of stats. I don't think we should add it here. Arguably we should have some way to get the config. Today we just say "read the config" file, but this PR makes a case that we should have a more active GetConfig() (we call it GetSpec() :) )

hqhq:
I just thought config is not always the same value as stats (or we call it spec), like memory, libcontainer should have a way to offer the real spec, so we don't have to deal with this additional on Docker side. And on Docker side, we also take memory as stats.

Anyway what you said also makes sense to me, you can close this PR if you think it's the best :)

vmarmol:
I will go ahead and close it so we can add a GetSpec() endpoint (as we have GetStats() today) which can give the isolation specification for the container.

PR #364, opened 2015-02-09
rename Set to Apply

hqhq:
The name `Set` would be used to do dymanic changes of resource configs
in the future. For now, `Apply` also makes more sense.

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -58,7 +58,7 @@ type subsystem interface {
 	// Removes the cgroup represented by 'data'.
 	Remove(*data) error
 	// Creates and joins the cgroup represented by data.
-	Set(*data) error
+	Apply(*data) error
 }
 
 type data struct {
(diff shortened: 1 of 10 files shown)


vmarmol:
LGTM

cgroups/fs/apply_raw.go
```diff
@@ -58,7 +58,7 @@ type subsystem interface {
 	// Removes the cgroup represented by 'data'.
 	Remove(*data) error
 	// Creates and joins the cgroup represented by data.
-	Set(*data) error
+	Apply(*data) error
```
dqminh: do you think `Join` is a better name since the comments on various places refers to the action as joining the cgroups ?

hqhq: `Join` is part of work,
`Apply` = `Join` + `Set`, so I think Apply would be better.

PS: I'll add `Set` interface and separate `Apply` to `Join` and `Set` in a subsequent PR.

hqhq:
ping.

dqminh:
LGTM

hqhq:
This patch is also included in https://github.com/docker/libcontainer/pull/376
So I'm gonna close this one, thanks for your reviews.

PR #367, opened 2015-02-09
API Refactoring 

crosbymichael:
This PR removes additional packages such as console, security, mount, network, etc moving the code and refactoring it into the root package.  This helps to avoid cyclic deps. 

This also moves Env, User, and Cwd from the container config to the Process.

This also has refactoring for the init and parent process handling so that it is more dry and does not have duplication where it's not needed.  

The other big change is that this moves the names of the veths from being generated by libcontainer to being generated at the level above.


linux_rootfs.go
```diff
@@ -0,0 +1,404 @@
...
+	if err := createDevices(config); err != nil {
+		return err
+	}
+	if err := setupPtmx(config); err != nil {
+		return err
+	}
+	// stdin, stdout and stderr could be pointing to /dev/null from parent namespace.
+	// Re-open them inside this namespace.
+	// TODO: Need to fix this for user namespaces.
+	if 0 == 0 {
```
mrunalp: Check for userns in config and skip for now? I think this is handled in #360

process.go
```diff
@@ -2,19 +2,28 @@ package libcontainer
...
+	// User will set the uid and gid of the executing process running inside the container
+	// local to the contaienr's user and group configuration.
+	User string
+
+	// Cwd will change the processes current working directory inside the container's rootfs.
+	Cwd string
+
 	// Stdin is a pointer to a reader which provides the standard input stream.
+	Stdin io.Reader
+
```
mrunalp: WDYT about also adding capabilities here? There are use cases where one would want to go and tweak a container with additional caps.

linux_init.go
```diff
@@ -0,0 +1,221 @@
...
+	Cwd    string          `json:"cwd"`
+	User   string          `json:"user"`
+	Config *configs.Config `json:"config"`
+}
+
+type initer interface {
+	Init() error
+}
+
+func newContainerInit(t initType, pipe *os.File) (initer, error) {
```
mrunalp: :+1: to separating types of processes.

configs/config.go
```diff
@@ -21,41 +21,34 @@ type Config struct {
...
 	// PivotDir allows a custom directory inside the container's root filesystem to be used as pivot, when NoPivotRoot is not set.
 	// When a custom PivotDir not set, a temporary dir inside the root filesystem will be used. The pivot dir needs to be writeable.
 	// This is required when using read only root filesystems. In these cases, a read/writeable path can be (bind) mounted somewhere inside the root filesystem to act as pivot.
 	PivotDir string `json:"pivot_dir,omitempty"`
 
-	// ReadonlyFs will remount the container's rootfs as readonly where only externally mounted
-	// bind mounts are writtable
-	ReadonlyFs bool `json:"readonly_fs,omitempty"`
+	// Path to a directory containing the container's root filesystem.
+	Rootfs string `json:"rootfs,omitempty"`
```
vmarmol: I would urge against changing the json field name since it will let us keep backwards compatibility. Otherwise it's gonna be really hard to get this rolled out in a new Docker :P

linux_setns_init.go
```diff
@@ -0,0 +1,33 @@
+// +build linux
```
vmarmol: Would it be possible to get all inits in the same package? If not, no big deal :) I know there are dependency issues.

crosbymichael: Do you mean a subpackage?

configs/config.go
crosbymichael: Well we are changing alot of things in this branch, moving things from config to process, etc.  It would be nice to make the changes that we want now and then not worry about it in the future.

I was thinking at the end i'll check in a `configs/v1` pkg that has the current config and write a function to migrate to and from the current config to our new struct.

vmarmol: SGTM

linux_setns_init.go
vmarmol: Yes

configs/config.go
crosbymichael: If you all want to make some changes to something now is the time ;)

linux_container.go
```diff
@@ -101,247 +86,109 @@ func (c *linuxContainer) Start(process *Process) (int, error) {
...
+}
+
+func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) *initProcess {
+	cloneFlags := c.config.Namespaces.CloneFlags()
+	if cloneFlags&syscall.CLONE_NEWUSER != 0 {
+		c.addUidGidMappings(cmd.SysProcAttr)
 		// Default to root user when user namespaces are enabled.
 		if cmd.SysProcAttr.Credential == nil {
 			cmd.SysProcAttr.Credential = &syscall.Credential{}
 		}
```
avagin: cmd.Env = append(cmd.Env, "_LIBCONTAINER_INITTYPE=userns")

linux_userns_sidecar_init.go
```diff
@@ -0,0 +1,35 @@
...
+func (l *linuxUsernsSideCar) Init() error {
+	if err := setupNetwork(l.config.Config); err != nil {
+		return err
+	}
+	if err := setupRoute(l.config.Config); err != nil {
+		return err
+	}
+	label.Init()
+	// InitializeMountNamespace() can be executed only for a new mount namespace
+	if l.config.Config.Namespaces.Contains(configs.NEWNET) {
```
avagin: I think we need to replace NEWNET on NEWNS here

linux_rootfs.go
```diff
@@ -0,0 +1,407 @@
...
+	if err := setupPtmx(config); err != nil {
+		return err
+	}
+	uid, err := config.HostUID()
+	if err != nil {
+		return err
+	}
+	// stdin, stdout and stderr could be pointing to /dev/null from parent namespace.
+	// Re-open them inside this namespace.
+	if uid == 0 {
```
avagin: Here was a comment "// FIXME: Need to fix this for user namespaces" and we check uid to detect USERNS. Currently we can use a proper check here:
if !config.Namespaces.Contains(configs.NEWUSER) {

mrunalp: +1

crosbymichael:
@avagin I took your suggestions and implemented randomly generating the temp veth name inside libcontainer and passing that state to the init process.

crosbymichael:
@avagin fixed, that was a typo on my part

avagin:
LGTM, thank you for this work.

vmarmol:
LGTM

PR #369, opened 2015-02-10
handle SIGCHLD when running as child subreaper

dqminh:
When running under child subreaper mode, it's useful for nsenter to be able to reap child processes. We have seen cases where spawned user processes were not reaped properly (https://github.com/creationix/nvm/issues/650)

I dont see how we can write a test case for this yet. Manual steps to reproduce this are:

```
# terminal 1
> docker pull soareschen/ubuntu-zsh
> docker run -it --rm --name test-nvm soareschen/ubuntu-zsh zsh
> source ~/.nvm/nvm.sh
> ps auxf # no defunct processes

# terminal 2
> docker exec -it test-nvm zsh
> source ~/.nvm/nvm.sh
> ps auxf # lots of defunct processes before this patch, and no defunct processes after
```

Thanks @soareschen for the test image

### DIFF
--- a/namespaces/nsenter/nsenter.c
+++ b/namespaces/nsenter/nsenter.c
@@ -12,6 +12,7 @@
 #include <string.h>
 #include <sys/prctl.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
 #include <getopt.h>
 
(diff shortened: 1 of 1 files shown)


vmarmol:
LGTM

mrunalp:
Thanks, looking good. I will quickly test it out.

mrunalp:
LGTM

PR #371, opened 2015-02-12
Add systemd support cpu.cfs_quota_us and cpu.cfs_period_us

coolljt0725:
Signed-off-by: Lei Jitang <leijitang@huawei.com>

`fs.Apply` support set `cpu.cfs_quota_us`, but `systemd.Apply` don't. This patch add systemd support   `cpu.cfs_quota_us`.

### DIFF
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -150,6 +150,12 @@ func Apply(c *cgroups.Cgroup, pid int) (map[string]string, error) {
 		return nil, err
 	}
 
+	// TODO: CpuQuota and CpuPeriod not available in systemd
+	// we need to manually join the cpu.cfs_quota_us and cpu.cfs_period_us
+	if err := joinCpu(c, pid); err != nil {
+		return nil, err
+	}
+
 	// -1 disables memorySwap
 	if c.MemorySwap >= 0 && c.Memory != 0 {
 		if err := joinMemory(c, pid); err != nil {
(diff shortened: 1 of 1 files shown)


cgroups/systemd/apply_systemd.go
```diff
@@ -150,6 +150,13 @@ func Apply(c *cgroups.Cgroup, pid int) (map[string]string, error) {
 		return nil, err
 	}
 
+	//TODO: CpuQuota is not available in systemd
+	if c.CpuQuota != 0 {
+		if err := joinCpu(c, pid); err != nil {
+			return nil, err
+		}
+	}
+
```
hqhq: where is CpuPeriod? We should support both of them.

coolljt0725:
update , add support cpu.cfs_period_us

cgroups/systemd/apply_systemd.go
```diff
@@ -196,6 +209,22 @@ func writeFile(dir, file, data string) error {
 	return ioutil.WriteFile(filepath.Join(dir, file), []byte(data), 0700)
 }
 
+func joinCpuQuota(c *cgroups.Cgroup, pid int) error {
```
mrunalp: I think it makes sense to have a single function for a single cgroup sub-system.

coolljt0725:
update as @mrunalp  suggestion, have a single function for cpu sub-system

coolljt0725:
ping @mrunalp

vmarmol:
LGTM

cgroups/systemd/apply_systemd.go
```diff
@@ -150,6 +150,12 @@ func Apply(c *cgroups.Cgroup, pid int) (map[string]string, error) {
 		return nil, err
 	}
 
+	// TODO: CpuQuota and CpuPeriod not available in systemd
```
dqminh: I think this is not a TODO.

rjnagal: This is still true that native systemd API doesn't support period and quota, right? We are using the filesystem path directly.

mrunalp: Yes, I think the comment is accurate.

mrunalp:
LGTM

PR #373, opened 2015-02-12
Use netlink to set hairpin mode

LK4D4:
This should fix problem with hairpin on EL6. I have no EL6 to  try though.
/cc @phemmer

### DIFF
--- a/netlink/netlink_linux.go
+++ b/netlink/netlink_linux.go
@@ -7,7 +7,6 @@ import (
 	"math/rand"
 	"net"
 	"os"
-	"path/filepath"
 	"sync/atomic"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


phemmer:
I just spun up a VM last time when I was digging into the issue. I can again, but someone with real world usage might be better.

LK4D4:
Actually I think this can be tested with `bridge` command from `iproute2`. Same netlink messages will be sent.

netlink/netlink_linux.go
```diff
@@ -1253,25 +1252,33 @@ func SetMacAddress(name, addr string) error {
...
 	if enabled {
-		writeVal = []byte("1")
-	} else {
-		writeVal = []byte("0")
+		mode[0] = byte(1)
 	}
-	if _, err := sysFile.Write(writeVal); err != nil {
+
+	br := newRtAttr(syscall.IFLA_PROTINFO|syscall.NLA_F_NESTED, nil)
+	newRtAttrChild(br, 4, mode)
```
mrunalp: The 4 here is IFLA_BRPORT_MODE, right? If so, could we define a constant for it?

LK4D4: Yeah, of course.

LK4D4:
@mrunalp I've added constant.

mrunalp:
@LK4D4 thanks, I will test this out today.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #377, opened 2015-02-13
Add a constant for the container console path.

mrunalp:

### DIFF
--- a/linux_console.go
+++ b/linux_console.go
@@ -12,6 +12,10 @@ import (
 	"github.com/docker/libcontainer/label"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


linux_console.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/docker/libcontainer/label"
 )
 
+const (
+	ContainerConsolePath string = "/dev/console"
```
crosbymichael: Should this be exported?

mrunalp: Ahh, yes, it shouldn't be. Fixed.

crosbymichael:
LGTM

PR #379, opened 2015-02-14
Add config generation for simple user namespace testing.

mrunalp:

### DIFF
--- a/nsinit/config.go
+++ b/nsinit/config.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"syscall"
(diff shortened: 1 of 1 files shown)


nsinit/config.go
```diff
@@ -70,6 +71,22 @@ func modify(config *configs.Config, context *cli.Context) {
 	config.AppArmorProfile = context.String("apparmor-profile")
 	config.ProcessLabel = context.String("process-label")
 	config.MountLabel = context.String("mount-label")
+
+	userns_uid := context.Int("userns-root-uid")
+	maxUid := 1 << 32
```
vmarmol: Is this meant to be an int32? This is undefined for them (shifting word size).

mrunalp: Yeah, want to set it to 2^32. Seems to be working but if it is undefined then need to change this. Thanks.

vmarmol: if we cast the 1 as an int64 it should be fine :)

mrunalp: Are you suggesting?
```
maxUid := int(int64(1) << 32)
```

vmarmol: Ah, well 2^32 is not a valid integer (int32). The largest unsigned integer is 2^32 -1. Should this be max unsigned int or max signed int? (both int32)

mrunalp: It is defined as unsigned uint32 in linux but is represented as int in golang.

mrunalp:
Updated to use math.MaxInt32.

vmarmol:
LGTM

PR #383, opened 2015-02-17
Linux has added a new capability audit_read

rhatdan:

### DIFF
--- a/vendor/src/github.com/syndtr/gocapability/capability/enum.go
+++ b/vendor/src/github.com/syndtr/gocapability/capability/enum.go
@@ -112,32 +112,33 @@ func (c Cap) String() string {
 		return "wake_alarm"
 	case CAP_BLOCK_SUSPEND:
 		return "block_suspend"
+	case CAP_AUDIT_READ:
+		return "audit_read"
 	}
 	return "unknown"
 }
 
+// POSIX-draft defined capabilities.
 const (
-	// POSIX-draft defined capabilities.
-
 	// In a system with the [_POSIX_CHOWN_RESTRICTED] option defined, this
 	// overrides the restriction of changing file ownership and group
 	// ownership.
-	CAP_CHOWN Cap = 0
+	CAP_CHOWN = Cap(0)
 
 	// Override all DAC access, including ACL execute access if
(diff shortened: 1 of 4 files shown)


vendor/src/github.com/syndtr/gocapability/capability/enum.go
```diff
@@ -112,6 +112,8 @@ func (c Cap) String() string {
 		return "wake_alarm"
 	case CAP_BLOCK_SUSPEND:
 		return "block_suspend"
+	case CAP_AUDIT_READ:
```
vmarmol: This is a change on the upstream dependency, we typically send a PR to that.

rhatdan: I sent one to them also.

vmarmol: Thanks :) this LGTM

mrunalp:
@rhatdan This looks good, but the right way to include this will be to have this merged upstream and then update the commit hash here https://github.com/docker/libcontainer/blob/master/update-vendor.sh#L46.

rhatdan:
Well I submitted it upstream.

LK4D4:
syndtr/gocapability#3 for reference

rhatdan:
Upstream has merged.

LK4D4:
@rhatdan Cool! Could you pls update revision in `update-vendor.sh`, run it and commit result here?

rhatdan:
@LK4D4 I would rather wait for @crosbymichael since there were other changes to upstream to stop this from happening again.  Basically allowing us to always drop caps even if the tool does not know about them.  Want to see how he wants to implement it.

rhatdan:
@mrunalp @crosbymichael Not sure how to handle this, since I have requested a pull from upstream gocapability.  

Should I run the update-vendor.sh and then send an updated patch. Or is this something you guys do?

This is why the tests are failing.

mrunalp:
@rhatdan you can update the commit id and run update-vendor.sh and then send the updated patch. Thanks.

mrunalp:
LGTM

PR #384, opened 2015-02-17
Introduce support for syscall filtering in containers #237

mheon:
Obsoletes #263

This introduces Seccomp system call filtering, which restricts system calls, preventing them from being made inside containers.

Compared to the previous pull request, several changes have been made:
- The previous version of the patches only allowed blacklists. This version can be configured to blacklist or whitelist system calls.
- Support for conditionally-blocked syscalls has been added. This allows a system call to be blocked only if its arguments meet certain conditions

The vendored bindings for libseccomp are less than optimal right now - they're located in the main libseccomp repository, so checking them out also checks out fifty-odd C files. If there are suggestions for a better way of doing things, I'd be open to changing this.


mrunalp:
@mheon The api branch was just merged into master. Would you mind rebasing? Thanks!

mheon:
@mrunalp Rebased, though it looks like I might have broken the test suite along the way - looking into it.

mrunalp:
@mheon Could you rebase again? Github is complaining about conflicts. I will test this out today.

mheon:
@mrunalp Rebased, unit tests passing.

Worth noting that there's a bug in libseccomp upstream at the moment which breaks conditional syscall blocking. A fix is incoming, but for the moment there are no tests on blocking a syscall based on multiple conditions

mrunalp:
@mheon Thanks!

mrunalp:
The seccomp tests aren't actually getting run in jenkins because of the seccomp tag.

sample_configs/seccomp.json
```diff
@@ -0,0 +1,225 @@
...
+            "gateway": "localhost",
+            "mtu": 1500,
+            "type": "loopback"
+        }
+    ],
+    "tty": true,
+    "user": "daemon",
+    "seccomp_config": {
+	"enable": true,
+	"whitelist": false,
```
mrunalp: whitelist_toggle?

mheon:
I've added the seccomp tag to the Dockerfile so the tests should be running automatically now. There were previously issues with running Seccomp tests on the CI server, but these seem to be resolved.

It should even be possible to remove the seccomp tag completely at this point, but it could still be desirable to build libcontainer without libseccomp support.

mheon:
Rebased on latest master. Some changes in integration tests, which broke on the new version of busybox.

mrunalp:
@mheon Thanks! I will try to get to this by end of today.

mrunalp:
@crosbymichael @LK4D4 ping

mrunalp:
I took a quick look and overall looks good.

LK4D4:
@mheon Can you rebase pls?

tiborvass:
@gabrtv here you go :)

mheon:
@LK4D4 Rebased

LK4D4:
@mheon I have question. Is this makes sense to release goseccomp as separate package under your name?

security/seccomp/types.go
```diff
@@ -0,0 +1,20 @@
...
+	ValueOne uint64 `json:"value_one"`
+	ValueTwo uint64 `json:"value_two,omitempty"`
+}
+
+type BlockedSyscall struct {
+	Name       string             `json:"name,"`
+	Conditions []SyscallCondition `json:"conditions,omitempty"`
+}
+
+type SeccompConfig struct {
```
LK4D4: I prefer call this type `Config` because `seccomp` part always obvious from package name.

security/seccomp/types.go
```diff
@@ -0,0 +1,20 @@
+package seccomp
+
+type SyscallCondition struct {
```
LK4D4: Could you write docstrings for this type? It is not obvious what it is.

security/seccomp/types.go
LK4D4: Here probably docstrings too. At least for `WhitelistToggle`

security/seccomp/seccomp.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+var (
+	actAllow libseccomp.ScmpAction = libseccomp.ActAllow
+	actDeny  libseccomp.ScmpAction = libseccomp.ActErrno.SetReturnCode(int16(syscall.EPERM))
+)
+
+// Filters given syscalls in a container, preventing them from being used
+// Started in the container init process, and carried over to all child processes
+func InitSeccomp(config SeccompConfig) error {
```
LK4D4: Config is pretty big and with arrays, better pass by pointer.

mheon:
@LK4D4 From an upstream perspective, it makes things easier for the Libseccomp maintainers. However, given it does create some issues with vendoring the bindings, I'd be willing to maintain a copy of the bindings independent of the main repository.

security/seccomp/seccomp.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"log"
+	"strings"
+	"syscall"
+
+	libseccomp "github.com/mheon/libseccomp/src/goseccomp"
+)
+
+var (
+	actAllow libseccomp.ScmpAction = libseccomp.ActAllow
```
LK4D4: `libseccomp.ScmpAction` is redundant, type will be inferred from right side.

security/seccomp/seccomp.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		if err = filter.AddArch(archConst); err != nil {
+			return fmt.Errorf("Error adding architecture %s to filter: %s", arch, err)
+		}
+	}
+
+	// Add a rule for each syscall
+	for _, call := range config.Syscalls {
+		if err = blockCall(config.WhitelistToggle, filter, call); err != nil {
```
LK4D4: You already know `defaultAction`, can you just pass it here instead of `WhitelistToggle`?

security/seccomp/seccomp.go
```diff
@@ -0,0 +1,141 @@
...
+		defaultAction = actAllow
+	}
+
+	filter, err := libseccomp.NewFilter(defaultAction)
+	if err != nil {
+		return fmt.Errorf("Error creating filter: %s", err)
+	}
+
+	// Unset no new privs bit
+	if err = filter.SetNoNewPrivsBit(false); err != nil {
```
LK4D4: I prefer to use masked errors in such places, like:
```
if err :=
```

security/seccomp/seccomp.go
```diff
@@ -0,0 +1,141 @@
+// +build linux,cgo,seccomp
```
LK4D4: Can we call this file `seccomp_linux.go`? And `unsupported.go` as `seccomp_unsupported.go`?

security/seccomp/types.go
```diff
@@ -0,0 +1,20 @@
...
+type BlockedSyscall struct {
+	Name       string             `json:"name,"`
+	Conditions []SyscallCondition `json:"conditions,omitempty"`
+}
+
+type SeccompConfig struct {
+	Enable          bool             `json:"enable"`
+	WhitelistToggle bool             `json:"whitelist_toggle"`
+	Architectures   []string         `json:"architectures,omitempty"`
+	Syscalls        []BlockedSyscall `json:"syscalls"`
```
LK4D4: `BlockedSyscall` should be pointer, because it contains slice.

standard_init_linux.go
```diff
@@ -77,6 +78,9 @@ func (l *linuxStandardInit) Init() error {
 	if err != nil {
 		return err
 	}
+	if err := seccomp.InitSeccomp(l.config.Config.SeccompConfig); err != nil {
```
LK4D4: I don't know, but I don't like idea that we just silently skip this step even if `l.config.Config.SeccompConfig` is not nil.

LK4D4: Also, will other processes in container inherit seccomp setup?

LK4D4:
@mheon Yeah, just looks not very well import path with `src` :) If upstream will maintain it, then okay. But if it will you, then I propose to create separate repo and just give link to it in upstream.

standard_init_linux.go
mheon: Children of container init will inherit automatically. Processes joined to the namespace (via exec, for example) will not inherit, so I have those execute seccomp_init() again to acquire the restrictions before joining the namespace.

security/seccomp/seccomp.go
mrunalp: +1

pcmoore:
Can someone please explain why including the libseccomp docker bindings with the rest of libseccomp is problematic?  I'm willing to make reasonable adjustments, but no one has explained the problem.

mheon:
@pcmoore All the Go dependencies for Libcontainer are included in this repository. The way of consistently acquiring or updating this externally-available dependencies is by a vendoring script, which clones their Git repositories. If the Libseccomp Go bindings are in the Libseccomp repository, this would grab the entire repository, including all the C source code and test suite. Ideally we'd just want the Go bindings. Right now, my solution is to clone the entire repo, and then remove every file that's not the Go bindings and license, which is kind of gross.

mrunalp:
@pcmoore @mheon Based on the libct backend dependencies (C + go) that we will end up pulling into libcontainer, I think it should be fine to pull all of libseccomp.
@LK4D4 @crosbymichael WDYT?

LK4D4:
@mrunalp I don't care.
Just don't like ugly imports and I'm not fan of libct bindings in repo too. I used to pure golang packages. Also I'm not sure about contributor/maintainer relationships in such projects, it isn't unusual that author of library is not willing to maintain all possible bindings in addition to library or giving write access to repo for all bindings authors.
Now you can see how much I don't care :)

mheon:
Requested changes have been made.

Most notably, we pull in the entire Libseccomp repository now, meaning the diffstat just exploded.

pcmoore:
Okay, so all is good now with respect to the repo?

rhatdan:
Certainly would be good to get this into docker-1.7.

diogomonica:
@rhatdan agreed. I think the big question is what a default profile will look like.

rhatdan:
@diogomonica I think Matt (And I) have come up with a pretty good default blocking profile.  Mostly stolen from other container tools and through suggestions from kernel developers.  If we could at least start getting some of this merged, then we could release an experimental patch to Rawhide, to see if anyone sees issues with the default list of blocked syscalls.

mrunalp:
@rhatdan There were some licensing concerns around merging this. (LGPL).

mheon:
@mrunalp We're looking into it. I wouldn't be opposed to relicensing, but the bindings are going in the upstream repository as well, so we want to examine options there. Updates on that (hopefully) soon!

On defaults: I think the most sane policy for a default blocklist is a blacklist, as I think we should allow syscalls introduced by new kernel versions by default, and in general whitelisting is more cumbersome than blacklisting for syscalls (an absolutely minimal whitelist to get from Docker to "return 0" is around 25 syscalls, and that goes up sharply in number if you want to do basic I/O, file operations, networking, etc).

hqhq:
Can we take a look at this PR https://github.com/docker/libcontainer/pull/529, it made it in another way.

pcmoore:
Interesting, it handles the seccomp BPF filter generation completely in Golang.  I'm obviously a little biased so I won't comment on which approach is better, but I will mention that just a quick glance at the code in PR #529 revealed a few issues/concerns:

* It doesn't appear that x32 is handled correctly, resulting in a gap in protection on some systems
* The syscall tables are wrong for ARM, e.g. OABI_SYSCALL_BASE (?)
* The code doesn't abstract away architecture specifics, e.g. socketcall()
* The generated BPF filter code looks like it might be twice as large as necessary resulting in unnecessary per-syscall performance overhead
* The generated BPF filter code has a trap/signal return action, yet I don't see a signal handler defined (?) ... possible this may be misunderstanding how Golang handles signals

I suspect if I look closer there are more issues as well, not to mention the long term maintenence issues (who will keep the syscall tables updated?).

rhatdan:
To me it would seem prudent to merge in libseccomp support and then allow the golang implementation to grow in support til it matches libseccomp.  At that point, I would suggest we look at this again.  But I really do not want this effort to get derailed.

diogomonica:
@rhatdan agreed. This said, BPF filter generation completely in Golang is totally the right way to go.

crosbymichael:
@mheon what about the license? Even if it's in upstream LGPL does not make sense for Go because of the static compiles.

mheon:
@crosbymichael I'd be fine with relicensing the bindings, though I'd have to go through @pcmoore to verify that he's alright with it as well.

Big issue from my perspective is that the actual libseccomp library, itself LGPL, is going to have to be statically-linked into Docker.

crosbymichael:
@mheon we can dynamically link it into docker but i still don't think it makes sense for Go because every go pkg is static.  what do you think?

rhatdan:
Why is this even a problem.  Don't we already staticly link in libc.so which is LGPL?

LK4D4:
@rhatdan I think no, we're not linking libc, that's why we have our own `user` library.

crosbymichael:
#613 was merged with a pure go implementation.  Thanks for your help and input on this feature.

rhatdan:
We would like to question this.  We have strong reservations about the go lang bindings.

mheon:
To add to this: I think the comments from @pcmoore on #529 provide a good summary of our reasoning for an alternative implementation (ease of maintenance, feature completeness, performance).

This PR hadn't seen much love in a long time (I'd moved on to other projects and not found time to rebase/solve the licensing issues). However, the way forward is fairly clear (relicense the bindings to a non-copyleft license... I was thinking BSD?), and I think I could have an updated version free of licensing hurdles by the end of tomorrow. I'd like this at least to be considered as an alternative before locking in to supporting a Golang implementation for the foreseeable future.

PR #385, opened 2015-02-18
Make usernamespaces work without sidecar process

crosbymichael:
VICTORY

After spending alot of time reading long strace's I finally found a way to get user namespaces with work with all of the standard code.

The big changes is trying to mknod a device, if you get an EPERM then bind mount it, done.

ping @mrunalp !!!!!


mrunalp:
Great, I got as far as figuring out that we needed to bind mount :)

linux_rootfs.go
```diff
@@ -209,6 +206,28 @@ func createDeviceNode(rootfs string, node *configs.Device) error {
 	if err := os.MkdirAll(parent, 0755); err != nil {
 		return err
 	}
+	if err := mknodDevice(dest, node); err != nil {
+		if os.IsExist(err) {
+			return nil
+		}
+		// containers running in a user namespace are not allowed to mknod
+		// devices so we can just bind mount it from the host.
```
mrunalp: could we pass a host bind flag here instead of relying on EPERM?

avagin:
Good job! Except to tiny comments, LGTM. Thanks.

crosbymichael:
@avagin updated

avagin:
LGTM, thanks

mrunalp:
LGTM

PR #386, opened 2015-02-19
Validation for user namespace in the config.

mrunalp:

### DIFF
--- a/configs/validate/config.go
+++ b/configs/validate/config.go
@@ -2,6 +2,7 @@ package validate
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/docker/libcontainer/configs"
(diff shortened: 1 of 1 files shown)


LK4D4:
Nice. Shouldn't we also check that there is no mappings in config too? Like we check for network namespace.

mrunalp:
@avagin @LK4D4 updated, thanks.

avagin:
LGTM, thanks

configs/validate/config.go
```diff
@@ -74,3 +78,16 @@ func (v *ConfigValidator) security(config *configs.Config) error {
 	}
 	return nil
 }
+
+func (v *ConfigValidator) usernamespace(config *configs.Config) error {
+	if config.Namespaces.Contains(configs.NEWUSER) {
+		if _, err := os.Stat("/proc/self/ns/user"); os.IsNotExist(err) {
```
dqminh: should we check if there's any other errors here too i.e. `err != nil` ?

mrunalp: I think it should be okay not to check for anything else here.

vmarmol:
LGTM

PR #390, opened 2015-02-21
nsenter: noop reference to C constructor

vbatts:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65134
Workaround suggested by iant

Reported here https://github.com/docker/docker/issues/9207#issuecomment-75354189

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/nsenter/nsenter.go
+++ b/nsenter/nsenter.go
@@ -10,3 +10,16 @@ void __attribute__((constructor)) init() {
 }
 */
 import "C"
+
+// AlwaysFalse is here to stay false
+// (and be exported so the compiler doesn't optimize out its reference)
+var AlwaysFalse bool
+
+func init() {
+	if AlwaysFalse {
+		// by referencing this C init() in a noop test, it will ensure the compiler
+		// links in the C function.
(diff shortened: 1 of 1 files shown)


vmarmol:
LGTM, although the built appears to be failing.

vbatts:
Interesting. Off hand it doesn't seem related.

avagin:
LGTM.

I think the test fails, because we use Cmd.Start(), but don't used Cmd.Wait(), so we don't wait until all cmd.goroutine are completed.

Cc: @crosbymichael

mrunalp:
@avagin Hmm, yeah, will have to call cmd.Wait instead of process.Wait. This will mean modifying the API to be able to call Wait on the cmd.

mrunalp:
Another option might be to use os.StartProcess.

crosbymichael:
I testing locally and it looks like that test races for some of the reasons above.  We can merge this one and look into the wait issues.

crosbymichael:
LGTM

PR #392, opened 2015-02-23
process: add Wait() and Pid() methods

avagin:
Currently we have a problem when buffers are used for std file
descriptors.  These buffers are filled from goroutines (Cmd.goroutine),
and we need to wait them to be sure that all data have been copied.

### DIFF
--- a/container.go
+++ b/container.go
@@ -5,8 +5,6 @@
 package libcontainer
 
 import (
-	"os"
-
 	"github.com/docker/libcontainer/configs"
 )
 
@@ -99,7 +97,7 @@ type Container interface {
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
 	// Systemerror - System error.
-	Start(process *Process) (pid int, err error)
+	Start(process *Process) (err error)
 
 	// Destroys the container after killing all running processes.
 	//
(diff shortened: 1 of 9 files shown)


dqminh:
I like the change, but should we change `linux_process.go` too ? The `wait` methods there (https://github.com/docker/libcontainer/blob/afa8443118347a1f909941aec2732039d28a9034/linux_process.go#L162) delegate to `os.Process#Wait` which doesnt actually wait for the goroutines in `exec.Cmd` to finish.
So if i understand this correctly, the bug will still be there (not waiting for all data in goroutines to be copied), although maybe harder to replicate with this change because we can now wait on the original process.

LK4D4:
@dqminh yup, I think you're right. We need additionally call `exec.Cmd.Wait()` for streams goroutines.

LK4D4:
@avagin WDYT about also moving `Signal` from `Container` to `Process`? Then we can treat container as process-runner and not `process-like` thing.

avagin:
@dqminh good catch. Thanks.
@LK4D4 It's a good idea about Signal(), I have added.

LK4D4:
@avagin nice, I also propose to remove `Signal` from `Container` to make all process in containers sorta equal from `libcontainer` point of view. Also I renamed files, so you need little rebase now.
Thanks!

mrunalp:
drone disappeared?

jfrazelle:
hang on let me see what i can do

mrunalp:
@jfrazelle thanks :)

jfrazelle:
may have to be force pushed :/ I tried resending the webhook but it thinks it already ran it

mrunalp:
Hmm, I can run the tests locally and update the thread.

mrunalp:
I did a few rounds of tests on Fedora and they are PASSING.

avagin:
@LK4D4 done

LK4D4:
LGTM mostly
but I noticed that this wait method for process with forks will be affected by #349 
And I have no idea how to fix this :/ Only calling `killCgroupProcesses` in `process.Wait` which seems weird.

mrunalp:
I think we can handle that as a separate issue. LGTM.

PR #394, opened 2015-02-23
Change os-prefix file naming to standard postfix naming

### LINKED ISSUE #393: File naming in project.
Now many files in repo called like `linux_*`. It is very confusing, because it is like standard `*_linux.go`, but not. Also tab completion is not very convenient with such naming.
What do you think about changing naming to standard postfix form?

LK4D4:
Fixes #393


crosbymichael:
LGTM

vmarmol:
LGTM

tianon:
Shouldn't `notify_test_linux.go` have been `notify_linux_test.go`?

PR #397, opened 2015-02-24
Use configs.NamespaceType as key for State.NamespacePathes

LK4D4:
I think this is more convenient because for working with namespaces we
using configs.NamespaceType.


crosbymichael:
LGTM

PR #403, opened 2015-02-25
Make possible to call config methods on values

LK4D4:
Because container.Config() returns values and you can't get pointer from
function call immediately. So it is impossible to call
container.Config().HostUID().


mrunalp:
LGTM

crosbymichael:
I don't understand the problem, go handles this fine.

http://play.golang.org/p/YKID4aieAT

crosbymichael:
nvm, i see what is happening, because of the return result

PR #409, opened 2015-02-26
rename test files so we can really test them

hqhq:
test files need to be named like *_test.go, a1c3e0db94579f59 mistakenly broken this, @LK4D4 .


crosbymichael:
nice catch

LGTM

PR #411, opened 2015-02-26
cgroups: add support for oom control

HuKeping:
This patch add support for diable OOM Killer.

Signed-off-by: Hu Keping <hukeping@huawei.com>


cgroups/fs/memory.go
```diff
@@ -49,6 +49,13 @@ func (s *MemoryGroup) Set(d *data) error {
 				return err
 			}
 		}
+
+		// The argument for memory.oom_control can only be 1 or 0
+		if d.c.OomKillDisable != 0 {
+			if err := writeFile(dir, "memory.oom_control", strconv.FormatInt(1, 10)); err != nil {
```
hqhq: Why not use `bool` instead?

HuKeping: Hum, I think bool is better

PR #412, opened 2015-02-26
cgroups: add support for oom control

HuKeping:
Redo #411


cgroups/fs/memory.go
```diff
@@ -26,7 +26,7 @@ func (s *MemoryGroup) Set(d *data) error {
 	}()
 
 	// Only set values if some config was specified.
-	if d.c.Memory != 0 || d.c.MemoryReservation != 0 || d.c.MemorySwap != 0 {
+	if d.c.Memory != 0 || d.c.MemoryReservation != 0 || d.c.MemorySwap != 0 || d.c.OomKillDisable {
```
vmarmol: Break out OomKillDisable into its own if statement. It shouldn't be bundled with these here (neither should Reservation and Memory for that matter, but that's out of scope :) ).

configs/cgroup.go
```diff
@@ -51,4 +51,7 @@ type Cgroup struct {
 
 	// Parent slice to use for systemd TODO: remove in favor or parent
 	Slice string `json:"slice"`
+
+	// Disable OOM Killer
```
vmarmol: nit: Whether to...

HuKeping:
why every time i redo the commit it will auto close the PR? Here's my steps:

$ git reset --soft HEAD^             (1)

$ edit                                      (2)

$ git commit -a -c ORIG_HEAD  (3)

$ git push -f 		              (4)

LK4D4:
@HuKeping I'm using `--amend` or `rebase -i` + `--amend` if I need to change not last commit. Seems like `github` have some protection against your case.

rjnagal:
I think your way makes the original commit id go away, which will close the
PR. Doing a rebase -i will preserve the id and github won't bother you.

On Thu, Feb 26, 2015 at 6:16 PM, Alexander Morozov <notifications@github.com

HuKeping:
get it,thanks

PR #415, opened 2015-02-26
Create cmd.StdinPipe and copy info to it from p.Stdin

LK4D4:
I'm not sure why it is work and just assignment not. But it's how it was
done in docker. Maybe because of io.Pipe instead of os.Pipe.

### DIFF
--- a/container_linux.go
+++ b/container_linux.go
@@ -5,6 +5,7 @@ package libcontainer
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -125,7 +126,17 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 		Path: c.initPath,
 		Args: c.initArgs,
 	}
-	cmd.Stdin = p.Stdin
+	if p.Stdin != nil {
+		stdin, err := cmd.StdinPipe()
+		if err != nil {
+			return nil, err
+		}
+		go func() {
+			io.Copy(stdin, p.Stdin)
+			stdin.Close()
+		}()
+	}
+
 	cmd.Stdout = p.Stdout
 	cmd.Stderr = p.Stderr
 	cmd.Dir = c.config.Rootfs


mrunalp:
LGTM

container_linux.go
```diff
@@ -125,7 +126,17 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
...
+		stdin, err := cmd.StdinPipe()
+		if err != nil {
+			return nil, err
+		}
+		go func() {
+			io.Copy(stdin, p.Stdin)
+			stdin.Close()
+		}()
+	}
+
```
dqminh: This should be the same as `cmd.Stdin = p.Stdin` except that in the original statement, `exec.Cmd` is doing it internally ( creating a os.Pipe and copied from cmd.Stdin ), so i'm not sure why the original didnt work. Do you have a test case ?

LK4D4: you can take my branch from https://github.com/LK4D4/docker/tree/new_libcontainer_api and integration tests will fail without that fix. I'll try to find author of original code in docker(`execdriver.StdConsole`), maybe he knows what's difference.

avagin:
@LK4D4 Would you describe a problem which is solved here?

container_linux.go
avagin: @LK4D4 In many cases it works fine without this hack, so I would prefer to use it only if it's required. For example if I set a terminal as Stdin, I expect that a process will be executed with the terminal and it will be able to call tty ioctls on the stdin descriptor. This patch breaks this ability.

LK4D4:
@avagin I don't know exact problem, but docker just not working without this :(

container_linux.go
LK4D4: You're right, I'll close pr.

avagin:
@LK4D4 Can we strace docker and investigate this problem? I can help with it on Monday.

LK4D4:
@avagin I'll try. Because I don't know how to implement that hack with current libcontainer API :)

PR #416, opened 2015-02-26
Don't join rootfs if path already prefixed by it

LK4D4:
In docker we evaluate all symlinks im path to check that they  not escaping
from rootfs and that gives us full path to mount.


rjnagal:
LGTM

mrunalp:
LGTM

PR #420, opened 2015-03-02
Implement stats for systemd

LK4D4:
It looks weird but works and not crashes application.


crosbymichael:
LGTM

I think this was just a bad rebase.

PR #421, opened 2015-03-03
Make NetworkInterface public

LK4D4:
In other case it is impossible to implement Stats() method from
interface outside libcontainer.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #422, opened 2015-03-03
Change process.Env type to map[string]string

LK4D4:


LK4D4:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html for reference
We're not sure that we want this.

dqminh:
> http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html for reference
We're not sure that we want this.

I think this probably works correcrly. What's your concern ?

mrunalp:
@dqminh Yes, it is correct per the standard, but the concern is that there might be legacy applications that may for some reason depend on the order.

LK4D4:
I'm closing this because actually we don't need write that code by our own.

PR #424, opened 2015-03-04
Fix panic when genericError constructor gets nil error

dqminh:
`genericError` constructor should not panic when we pass in nil error, but we should not pass nil error actually. This fixes it.


LK4D4:
LGTM

mrunalp:
LGTM

PR #425, opened 2015-03-04
cgroups: use Set instead of Apply in Freeze

hqhq:
So Apply function of freezer can be as sample as other subsystems.


crosbymichael:
LGTM

mrunalp:
LGTM

PR #426, opened 2015-03-04
Add the file close operation before function return to release resource

MabinGo:
In function notifyOnOOM,it do the open file operations but do not be close before return,so close file operations should be added to release resource.


crosbymichael:
LGTM

thanks!

PR #427, opened 2015-03-04
A few minor fixes

avagin:


LK4D4:
LGTM
But jenkins is not working :)

crosbymichael:
Looks like you need to gofmt some files

avagin:
@crosbymichael @LK4D4  thanks. I have uploaded a fixed version.

process.go
```diff
@@ -54,8 +54,12 @@ func (p Process) Wait() (*os.ProcessState, error) {
 
 // Pid returns the process ID
 func (p Process) Pid() (int, error) {
+	// A random negative values can be dengerous, because
+	// they can be valid for the kill system call.
+	const errorPid = -(int(^uint(0)>>1) - 1) // MinInt
```
mrunalp: Can we use a constant from math package here?

avagin:
@mrunalp Done. Thanks.

mrunalp:
The kernel seems to have a max limit of 2^22 - 1 = 4194303 for pid_max. There couldn't be more process groups than that so this should actually be safe as math.MinInt32 is -2147483648

dqminh:
> The kernel seems to have a max limit of 2^22 - 1 = 4194303 for pid_max. There couldn't be more process groups than that so this should actually be safe as math.MinInt32 is -2147483648

Yes i think you are correct. TIL that :heart: 
(https://github.com/torvalds/linux/blob/master/include/linux/threads.h#L33 )

LGTM

mrunalp:
LGTM

PR #429, opened 2015-03-05
fix apply error when we not mount cpu subsystem

hqhq:
Find cgroup mountpoint dir through a specific subsystem is not reliable,
we don't know which subsystem users will or will not mount, we can't
assume that, only we can assume is that users mount cgroup subsystems
on the same dir.

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -54,12 +54,10 @@ func getCgroupRoot() (string, error) {
 		return cgroupRoot, nil
 	}
 
-	// we can pick any subsystem to find the root
-	cpuRoot, err := cgroups.FindCgroupMountpoint("cpu")
+	root, err := cgroups.FindCgroupMountpointDir()
 	if err != nil {
 		return "", err
 	}
-	root := filepath.Dir(cpuRoot)
 
 	if _, err := os.Stat(root); err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


crosbymichael:
ping @vmarmol @rjnagal

rjnagal:
We always assume that cpu cgroup is required, so this is not a problem in practice.
But all cgroups are currently mounted under the same dir on most distro, so the new logic would work too. ( I briefly looked up the named ones like systemd on some machines and they are under the same dir).

I don't think the change is necessary, but it doesn't break anything.

vmarmol:
LGTM

vmarmol:
It is no worse than what we have today. It will handle if a user doesn't mount CPU, but it will fail if a user mounts cgroups in different directories.

hqhq:
@rjnagal assuming cpu cgroup is required is too specific, at least we need to check it somewhere or add this in doc, but that's still not good enough. And many of clients, they only use cpuset and memory cgroup in produces, it's likely they won't mount cpu cgroup or other cgroups they don't want to use. I did see that many times.

Cgroups mounted under the same dir, that's also assumed before this PR, all subsystems use the same `d.root`, if you want to use Docker, you need to mount cgroup, and you mount it in the way all other distros did (under the same dir), that's common and reasonable.

So I think this PR is progress and necessary, what do you think?

dqminh:
LGTM

vmarmol:
@rjnagal I'll wait for your LGTM, but this should be safe :)

rjnagal:
LGTM. This change is fine. I did think that we mentioned cpu hierarchy being required, but @hqhq is right in that we didn't mention it anywhere.

PR #433, opened 2015-03-06
use real cgroup files for cgroup tests

hqhq:
Right now, we test cgroup features and apis with normal files, not cgroup files managed by cgroupfs, that is not reasonable.

One problem is that stats tests write a lot of files which are not allowed to be written in cgroup files, these can't be changed easily, we need to figure out a better way to test stats.

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -231,6 +231,12 @@ func (raw *data) parent(subsystem string) (string, error) {
 }
 
 func (raw *data) path(subsystem string) (string, error) {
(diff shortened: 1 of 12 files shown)


vmarmol:
I think there is definitely an argument for tests that actually affect the filesystem, but I think those are integration tests, not unit tests. There is value in having these smaller tests that you don't need to depend on anything in the machine.

hqhq:
@vmarmol Make sense, let's leave these to integration tests, I'm gonna close this PR and move some fix in this PR to https://github.com/docker/libcontainer/pull/437.

PR #437, opened 2015-03-09
fix some cgroups issues

hqhq:
1. add subsystem mounted check in `d.path()`, so we can really get `IsNotFound` error and skip `Set`.
2. cgroup unit test use temp path, not real cgroup path, so we should create the path by `filepath.Join()` instead of `d.path()`.

### DIFF
--- a/cgroups/fs/devices_test.go
+++ b/cgroups/fs/devices_test.go
@@ -25,7 +25,7 @@ func TestDevicesSetAllow(t *testing.T) {
 	defer helper.cleanup()
 
 	helper.writeFileContents(map[string]string{
-		"device.deny": "a",
+		"devices.deny": "a",
 	})
 
 	helper.CgroupData.c.AllowAllDevices = false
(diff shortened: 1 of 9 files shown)


cgroups/fs/devices_test.go
```diff
@@ -25,7 +25,7 @@ func TestDevicesSetAllow(t *testing.T) {
 	defer helper.cleanup()
 
 	helper.writeFileContents(map[string]string{
-		"device.deny": "a",
+		"devices.deny": "a",
```
vmarmol: Nice catch :) I've been meaning to const all these paths and use those consts.

cgroups/fs/util_test.go
```diff
@@ -31,12 +31,12 @@ func NewCgroupTestUtil(subsystem string, t *testing.T) *cgroupTestUtil {
...
 		c: &configs.Cgroup{},
 	}
-	tempDir, err := ioutil.TempDir("", fmt.Sprintf("%s_cgroup_test", subsystem))
+	tempDir, err := ioutil.TempDir("", "cgroup_test")
 	if err != nil {
 		t.Fatal(err)
 	}
 	d.root = tempDir
-	testCgroupPath, err := d.path(subsystem)
+	testCgroupPath := filepath.Join(d.root, subsystem)
```
vmarmol: nice! This used to not be possible, thanks for the fix.

vmarmol:
LGTM

crosbymichael:
LGTM

tianon:
Went to go fix the regression in #337 (essentially making "blkio" required) and found that this PR already did so. :+1: :heart: :heart_eyes:

PR #439, opened 2015-03-10
nsenter: fix the -Wunused-variable warning

vbatts:
The change 699429e60f23ab0fa3bdd97b6326316be08791ad that allows the
constructor attribute to not be optimized out on gccgo has resulted in a
warning when compiling on the golang compiler.

```
./nsenter.go: In function ‘_cgo_51505a0edd5d_Cfunc_init’:
./nsenter.go:40:49: warning: unused variable ‘a’ [-Wunused-variable]
```

the generated code produced an unused struct like:

```
void
_cgo_d6cfae95ae01_Cfunc_init (void *v)
{
  struct
  {
    char unused;
  } __attribute__ ((__packed__, __gcc_struct__)) * a = v;
  init ();
}
```

Truly the "fix" would be upstream in cgo. If it knows it is producing an
unused struct, then it should also include __attribute__ ((unused))

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/nsenter/nsenter.go
+++ b/nsenter/nsenter.go
@@ -1,25 +1,12 @@
-// +build linux
+// +build linux,!gccgo
 
 package nsenter
 
 /*
 #cgo CFLAGS: -Wall
 extern void nsexec();
-void __attribute__((constructor)) init() {
+void __attribute__((constructor)) init(void) {
 	nsexec();
 }
 */
 import "C"
-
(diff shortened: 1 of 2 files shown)


vbatts:
ping @crosbymichael

vmarmol:
LGTM, thanks! I was seeing this recently.

mrunalp:
LGTM

PR #441, opened 2015-03-11
add parameter to Set api

hqhq:
We only have libcontainer.Container on Docker side, can't change `config`
in linuxContainer, pass config to libcontainer so we can change config of
container.

### DIFF
--- a/container.go
+++ b/container.go
@@ -96,7 +96,7 @@ type Container interface {
 	//
 	// errors:
 	// Systemerror - System error.
-	Set() error
+	Set(config configs.Config) error
 
 	// Start a process inside the container. Returns error if process fails to
 	// start. You can track process lifecycle with passed Process structure.
--- a/container_linux.go
+++ b/container_linux.go
@@ -78,9 +78,10 @@ func (c *linuxContainer) Stats() (*Stats, error) {
 	return stats, nil
 }
 
-func (c *linuxContainer) Set() error {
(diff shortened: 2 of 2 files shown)


container.go
```diff
@@ -96,7 +96,7 @@ type Container interface {
 	//
 	// errors:
 	// Systemerror - System error.
-	Set() error
+	Set(config configs.Config) error
```
dqminh: This gives me the assumption that the API allows us to change any of the configuration, while it probably only makes sense to change the cgroup limits (and possibly rlimits ).

Do you think we should define a subset of configuration (i.e. cgroup values, rlimits etc. ) that can be changed for a running container instead of reusing `Config` here ?

mrunalp: I think Set could have more functionality in future to cover more than cgroups, so taking in the entire config makes sense to me.

hqhq: Yes, I was intend to make Set more functionality, not just change resource configs. We can reusing Config in libcontainer, and leave it to Docker to decide which Config can be changed.

hqhq:
Hi, Set command from Docker side is out, https://github.com/docker/docker/issues/11442 
that PR needs this commit, can you help review? Thanks.

mrunalp:
LGTM @crosbymichael @vmarmol ping

dqminh:
LGTM

vmarmol:
LGTM

PR #442, opened 2015-03-11
Mounting a tmpfs directory needs to inherit directory permissions from base

rhatdan:

### DIFF
--- a/rootfs_linux.go
+++ b/rootfs_linux.go
@@ -100,7 +100,23 @@ func mount(m *configs.Mount, rootfs, mountLabel string) error {
 			return err
 		}
 		return syscall.Mount(m.Source, dest, m.Device, uintptr(m.Flags), "")
-	case "tmpfs", "mqueue", "devpts", "sysfs":
+	case "tmpfs":
+		stat, err := os.Stat(dest)
+		if err != nil {
+			if err := os.MkdirAll(dest, 0755); err != nil && !os.IsExist(err) {
+				return err
+			}
+		}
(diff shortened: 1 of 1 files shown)


rootfs_linux.go
```diff
@@ -100,7 +103,26 @@ func mount(m *configs.Mount, rootfs, mountLabel string) error {
 			return err
 		}
 		return syscall.Mount(m.Source, dest, m.Device, uintptr(m.Flags), "")
-	case "tmpfs", "mqueue", "devpts", "sysfs":
+	case "tmpfs":
+		if stat, err = os.Stat(dest); err == nil {
```
dqminh: This should be `err != nil` right ? because we only want to create the directory if it's not there.

rootfs_linux.go
```diff
@@ -100,7 +103,26 @@ func mount(m *configs.Mount, rootfs, mountLabel string) error {
...
+			return err
+		}
+		if stat != nil {
+			if file, err = os.Open(dest); err != nil {
+				return err
+			}
+			defer file.Close()
+			if err = file.Chmod(stat.Mode()); err != nil {
+				return err
+			}
```
dqminh: can we use `os.Chmod(dest, stat.Mode())` here instead of having to open the file and use fchmod ?

rhatdan:
Cleaned up patch and fixed mkdirall call.

crosbymichael:
LGTM

dqminh:
LGTM

PR #444, opened 2015-03-11
Don't label mqueue when mounting

ncdc:
mqueue shouldn't be labeled when mounting; otherwise, the mount will
fail.

Signed-off-by: Andy Goldstein <agoldste@redhat.com>


crosbymichael:
LGTM

mrunalp:
LGTM

PR #445, opened 2015-03-11
Add information Type method for Factory

LK4D4:
I need this for `docker info` don't want to use type assertions because actually `LinuxFactory` should be private.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #450, opened 2015-03-16
Fix path to /dind

avagin:
That script was moved a few days ago
https://github.com/docker/docker/pull/11373


tianon:
:+1: LGTM

dqminh:
LGTM

tianon:
Do you think it might be better to switch `master` to an explicit commit reference to help insulate against issues like this in the future?

mrunalp:
LGTM

mrunalp:
@tianon I think it maybe good to just pull the latest. If we hit this again, we can revisit using commit id :)

tianon:
:+1:

PR #451, opened 2015-03-16
mount: sysfs also doesn't need to be labelled like mqueue.

mrunalp:


mrunalp:
We did not label sysfs earlier and it is causing selinux issues. ping @LK4D4 @crosbymichael @vmarmol

vmarmol:
LGTM

PR #453, opened 2015-03-16
Do not fail cgroups setup if parent cgroup does not exist.

vishh:
As of now if the parent cgroup path is absolute, we expect the cgroups to exist. If it is not absolute, everything works as expected.

### DIFF
--- a/cgroups/fs/cpuset.go
+++ b/cgroups/fs/cpuset.go
@@ -17,12 +17,9 @@ type CpusetGroup struct {
 func (s *CpusetGroup) Apply(d *data) error {
 	dir, err := d.path("cpuset")
 	if err != nil {
-		if cgroups.IsNotFound(err) {
-			return nil
-		} else {
-			return err
-		}
+		return err
 	}
+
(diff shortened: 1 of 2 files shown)


mrunalp:
Build failed -- exec_test.go:297: [8] System error: mountpoint for devices not found

cgroups/fs/cpuset.go
```diff
@@ -17,12 +17,9 @@ type CpusetGroup struct {
 func (s *CpusetGroup) Apply(d *data) error {
 	dir, err := d.path("cpuset")
 	if err != nil {
-		if cgroups.IsNotFound(err) {
-			return nil
-		} else {
-			return err
-		}
+		return err
```
vmarmol: Why this change only in cpuset?

vishh: The rest of the cgroup setups use data.path() and cpuset does something different. This change makes sure it stays in sync.

vishh:
Fixed the cause for the test failure. PTAL.

vmarmol:
LGTM

PR #457, opened 2015-03-18
Prefer syscall.Kill over os.Process.Signal

ibuildthecloud:
In go 1.4 os.Process.Signal started masking the syscall.ESRCH error
with its own error.  Other code in Docker looks checks for ESRCH to
handle situations in which a process has just died.  By using
syscall.Kill we can communicate back the proper error.

Reference to go code where it masks: https://github.com/golang/go/blob/ecd630de6dd2e5285f18c650207b5c84649402db/src/os/exec_unix.go#L54

### DIFF
--- a/process_linux.go
+++ b/process_linux.go
@@ -45,7 +45,7 @@ func (p *setnsProcess) startTime() (string, error) {
 }
 
 func (p *setnsProcess) signal(s os.Signal) error {
-	return p.cmd.Process.Signal(s)
+	return maybeSyscall(p.cmd.Process, s)
 }
 
 func (p *setnsProcess) start() (err error) {
@@ -236,5 +236,14 @@ func (p *initProcess) createNetworkInterfaces() error {
 }
 
 func (p *initProcess) signal(s os.Signal) error {
-	return p.cmd.Process.Signal(s)
+	return maybeSyscall(p.cmd.Process, s)
+}
+
+func maybeSyscall(process *os.Process, s os.Signal) error {
+	if sys, ok := s.(syscall.Signal); ok {
+		// Prefer syscall.Kill because os.Process.Signal masks ESRCH
+		return syscall.Kill(process.Pid, sys)
+	} else {
+		return process.Signal(s)
+	}
 }


process_linux.go
```diff
@@ -236,5 +236,14 @@ func (p *initProcess) createNetworkInterfaces() error {
...
+	return maybeSyscall(p.cmd.Process, s)
+}
+
+func maybeSyscall(process *os.Process, s os.Signal) error {
+	if sys, ok := s.(syscall.Signal); ok {
+		// Prefer syscall.Kill because os.Process.Signal masks ESRCH
+		return syscall.Kill(process.Pid, sys)
+	} else {
+		return process.Signal(s)
+	}
```
dqminh: I think using syscall.Kill as-is is unsafe as process.Signal performs several useful safety check ( make sure that we dont do `kill -1` or `kill -0` or not having to perform `kill` at all if it's certain that the process is finished ). We at least have to check for PID -1/0 here.

What do you think about checking error string coming out of `process.Signal` (either here or in the caller) i.e. checking for `error: process already finished/released...` so we dont have to duplicate `process.Signal` logic ?

ibuildthecloud: The only safety checks I see is that ensure that the PID is not 0 or 1.  I think in libcontainer we should be pretty safe here as I can't think of a flow of how we could have launched a process with PID 0 or 1 short of the caller setting the PID value in the struct.

LK4D4: @dqminh It returns private error on `ESRCH`
http://golang.org/src/os/exec_unix.go#L37

process_linux.go
```diff
@@ -236,5 +236,14 @@ func (p *initProcess) createNetworkInterfaces() error {
 }
 
 func (p *initProcess) signal(s os.Signal) error {
-	return p.cmd.Process.Signal(s)
+	return maybeSyscall(p.cmd.Process, s)
+}
+
+func maybeSyscall(process *os.Process, s os.Signal) error {
```
crosbymichael: Why is this func needed at all?  Why can we not use "one true way"?

ibuildthecloud: Not sure if I follow.  I'd prefer we just always use syscall.Kill but because of the type assertion I fell back to process.Signal.  Is it possible to get a os.Signal that is not a syscall.Signal?

LK4D4: `os.Signal` is interface and `syscall.Singnal` its implementation

ibuildthecloud: @crosbymichael @LK4D4 Where do we stand on this.  Is there some safe way that I could always just call syscall.Kill and never call process.Signal

crosbymichael: Just do syscall.Kill always so no need for this func

process_linux.go
dqminh: @ibuildthecloud I looked at the code again and i think you are correct. It should be safe to say that we cant signal an unstarted process ( i.e. process with invalid PID ). So i think it's pretty safe to always call `syscall.Kill`

PR #459, opened 2015-03-19
Add TmpfsRoot option

LK4D4:


mrunalp:
Needs rebase?

LK4D4:
@mrunalp Rebased

factory_linux.go
```diff
@@ -10,7 +10,9 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"syscall"
 
+	"github.com/docker/docker/pkg/mount"
```
mrunalp: Do we need to vendor this?

crosbymichael: Humm, i forgot about this.  Maybe we want to keep it in docker so that we don't have to add this dep.  We can define this same function there since that is the reason why we have these.

LK4D4: We using it already for `GetMounts` mostly.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #462, opened 2015-03-20
Add `Cleanup()` to Factory interface

cpuguy83:
Cleanup releases all resources held by the given factory.
In the case of the LinuxFactory it cleans up the `tmpfs` mount.


LK4D4:
Hehe, you were faster :) #463

crosbymichael:
I think destroy sounds better

PR #463, opened 2015-03-20
Destroy method for Factory

LK4D4:
We probably need this, because now we have mounted dir after docker stops.

### DIFF
--- a/factory_linux.go
+++ b/factory_linux.go
@@ -192,6 +192,18 @@ func (l *LinuxFactory) Type() string {
 	return "libcontainer"
 }
 
+func (l *LinuxFactory) Destroy() error {
+	ls, err := ioutil.ReadDir(l.Root)
+	if err != nil {
+		return newGenericError(err, SystemError)
+	}
+	if len(ls) != 0 {
+		return newGenericError(fmt.Errorf("%s is not empty", l.Root), RootIsNotEmpty)
+	}
+	syscall.Unmount(l.Root, syscall.MNT_DETACH)
+	return os.Remove(l.Root)
+}
+
 // StartInitialization loads a container by opening the pipe fd from the parent to read the configuration and state
 // This is a low level implementation detail of the reexec and should not be consumed externally
 func (l *LinuxFactory) StartInitialization(pipefd uintptr) (err error) {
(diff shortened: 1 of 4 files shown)


factory_linux.go
```diff
@@ -192,6 +192,18 @@ func (l *LinuxFactory) Type() string {
...
 
+func (l *LinuxFactory) Destroy() error {
+	ls, err := ioutil.ReadDir(l.Root)
+	if err != nil {
+		return newGenericError(err, SystemError)
+	}
+	if len(ls) != 0 {
+		return newGenericError(fmt.Errorf("%s is not empty", l.Root), RootIsNotEmpty)
+	}
+	syscall.Unmount(l.Root, syscall.MNT_DETACH)
```
crosbymichael: I guess this is still kinda weird because mounting is optional for the root.

LK4D4: But you said yourself that I don't need to check if it's mounted :)

crosbymichael: Ya, but why is unmount even here at all?  shouldn't it be configurable and only called if you passed the option?  Like we need to register destroy funcs that are called at destroy time?

LK4D4: But why not if it will just error on not mounted root. With destroy functions order will matter, but I can try to write them if you want.

avagin:
Why do we need to mount root from libcontainer?

LK4D4:
@avagin Why not? Root's nature shouldn't make sense to anything apart from Factory implementation.

avagin:
@LK4D4 I think it isn't a common practice. Usually applications use /var/run for such sort of data.

factory_linux.go
mrunalp: I think we should try to unmount only if we mounted it.

avagin: We may create .libcontainer_mount file after mounting root, and then umount root only if this file exists.

crosbymichael:
Maybe we should just use /var/run for this stuff

mrunalp:
Should the mounting root to tmpfs be moved to docker like @avagin suggested? That way cleanup could also be done there and libcontainer wouldn't have to care about how rootfs is mounted.

LK4D4:
Close because I don't like it too.

PR #464, opened 2015-03-20
Throw an error if cgroup tries to set cpu-shares more/less than the maximum/minimum permissible value.

shishir-a412ed:
If I pass a value for `cpu-shares` larger than the maximum allowed value, the kernel just changes it to the maximum allowed value. This is confusing for the user as what he requested and what is set is different.

e.g. if the max possible value is 262144 and I tried to set cpu-shares as 280000, 
This code will throw an application error saying "The maximum allowed cpu-shares is 262144".

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -1,8 +1,11 @@
 package fs
 
 import (
+	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"sync"
@@ -75,9 +78,12 @@ type data struct {
 }
 
 func (m *Manager) Apply(pid int) error {
+
 	if m.Cgroups == nil {
 		return nil
 	}
+
+	var c = m.Cgroups
 
 	d, err := getCgroupData(m.Cgroups, pid)
 	if err != nil {
@@ -107,6 +113,28 @@ func (m *Manager) Apply(pid int) error {
 		paths[name] = p
 	}
 	m.Paths = paths
+
+	var cpuShares int64
+
+	fd, err := os.Open(path.Join(m.Paths["cpu"], "cpu.shares"))
+	if err != nil {
+		return err
+	}
+
+	_, err = fmt.Fscanf(fd, "%d", &cpuShares)
+	if err != nil && err != io.EOF {
+		return err
+	}
+
+	fd.Close()
+
+	if c.CpuShares != 0 {
+		if c.CpuShares > cpuShares {
+			return fmt.Errorf("The maximum allowed cpu-shares is %d", cpuShares)
+		} else if c.CpuShares < cpuShares {
+			return fmt.Errorf("The minimum allowed cpu-shares is %d", cpuShares)
+		}
+	}
 
 	return nil
 }
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -5,8 +5,10 @@ package systemd
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -240,6 +242,28 @@ func (m *Manager) Apply(pid int) error {
 	}
 
 	m.Paths = paths
+
+	var cpuShares int64
+
+	fd, err := os.Open(path.Join(m.Paths["cpu"], "cpu.shares"))
+	if err != nil {
+		return err
+	}
+
+	_, err = fmt.Fscanf(fd, "%d", &cpuShares)
+	if err != nil && err != io.EOF {
+		return err
+	}
+
+	fd.Close()
+
+	if c.CpuShares != 0 {
+		if c.CpuShares > cpuShares {
+			return fmt.Errorf("The maximum allowed cpu-shares is %d", cpuShares)
+		} else if c.CpuShares < cpuShares {
+			return fmt.Errorf("The minimum allowed cpu-shares is %d", cpuShares)
+		}
+	}
 
 	return nil
 }


cgroups/systemd/apply_systemd.go
```diff
@@ -212,6 +213,17 @@ func (m *Manager) Apply(pid int) error {
 
 	m.Paths = paths
 
+	cpuShares, err := ioutil.ReadFile(path.Join(m.Paths["cpu"], "cpu.shares"))
+	if err != nil {
+		return err
+	}
+
+	if c.CpuShares != 0 {
+		if strings.TrimSpace(strconv.FormatInt(c.CpuShares, 10)) != strings.TrimSpace(string(cpuShares)) {
```
LK4D4: I think it's better to use `fmt.Sprintf("%d", c.CpuShares)`, also you don't need `TrimSpace` for this.
But probably it's better to `Fscanf` int directly from file. I'm not sure if it makes sense for one value, but sometimes you need to wrap opened file to `BufReader` for `Fscanf`

shishir-a412ed: yeah I used TrimSpace because even though the values {1: Set by the kernel, 2: Passed by docker run} were same ... the leading and trailing spaces were failing the comparisons.

shishir-a412ed: I just tried, `fmt.Sprintf()`, It is printing nothing on docker daemon logs or client, which I was expecting. we need to throw the error back from libcontainer to docker. Let me know if this works for you.

I ll take a look at FScanf and see if I can read `in64` directly from the file.

LK4D4: `fmt.Sprintf()` creates string, not printing it.

shishir-a412ed: @LK4D4 Something like this ?

cgroups/systemd/apply_systemd.go
```diff
@@ -212,6 +213,17 @@ func (m *Manager) Apply(pid int) error {
...
 	m.Paths = paths
 
+	cpuShares, err := ioutil.ReadFile(path.Join(m.Paths["cpu"], "cpu.shares"))
+	if err != nil {
+		return err
+	}
+
+	if c.CpuShares != 0 {
+		if fmt.Sprintf("%d", c.CpuShares) != strings.TrimSpace(string(cpuShares)) {
+			return fmt.Errorf("The maximum allowed cpu-shares is %s\n", string(cpuShares))
```
LK4D4: `\n` at the end is not needed
You don't need `string`, because `fmt` will do this for you.

cgroups/systemd/apply_systemd.go
```diff
@@ -212,6 +213,17 @@ func (m *Manager) Apply(pid int) error {
 
 	m.Paths = paths
 
+	cpuShares, err := ioutil.ReadFile(path.Join(m.Paths["cpu"], "cpu.shares"))
```
LK4D4: Probably better to create temp variable for `strings.TrimSpace(string(cpuShares))` and then use it in condition and error

vmarmol:
This won't trigger for only max, it will happen for min too. For example, try to write 1, the kernel will cap it at 2.

We should also be doing this in fs too no?

hqhq:
The minimum number is mentioned in docs, the maximum isn't.
But should this be put to Docker side, where we do all other resource configs check? Right now it's in `daemon/create.go`. In libcontainer, we deal with cgroupfs directly, and can leave the checking to cgroupfs.

shishir-a412ed:
@LK4D4 I have cleaned up the code. Please review it and let me know if any more changes are required.

@vmarmol I have update the code to handle both maximum and minimum checks.

cgroups/systemd/apply_systemd.go
```diff
@@ -212,6 +214,28 @@ func (m *Manager) Apply(pid int) error {
...
+		return err
+	}
+
+	fd.Close()
+
+	if c.CpuShares != 0 {
+		if c.CpuShares > cpuShares {
+			return fmt.Errorf("The maximum allowed cpu-shares is %d\n", cpuShares)
+		} else if c.CpuShares < cpuShares {
+			return fmt.Errorf("The minimum allowed cpu-shares is %d\n", cpuShares)
```
dqminh: i think you dont need `\n` here

dqminh:
@shishir-a412ed I think we should do this check in `cgroup/fs` too

cgroups/systemd/apply_systemd.go
shishir-a412ed: @dqminh removed \n.

shishir-a412ed:
@vmarmol @dqminh @LK4D4 
Just to confirm when you say you want the same checks replicated in `cgroups/fs` too.

You want me to put the same checks in this function ?
https://github.com/docker/libcontainer/blob/master/cgroups/fs/apply_raw.go#L77

Shishir

LK4D4:
@dqminh +1
It is actually same code for fs and systemd if we know path to cgroup. Probably we need to work more abstractions around paths.

shishir-a412ed:
@LK4D4 @vmarmol I tried to put the same change in https://github.com/docker/libcontainer/blob/master/cgroups/fs/apply_raw.go#L77, however it is failing.

Can you please elaborate where the change for cgroups/fs needs to be done.
Or can we just merge this PR with systemd change ?

Shishir

cgroups/fs/apply_raw.go
```diff
@@ -108,6 +110,28 @@ func (m *Manager) Apply(pid int) error {
...
+	fd.Close()
+
+	if c.CpuShares != 0 {
+		if c.CpuShares > cpuShares {
+			return fmt.Errorf("The maximum allowed cpu-shares is %d", cpuShares)
+		} else if c.CpuShares < cpuShares {
+			return fmt.Errorf("The minimum allowed cpu-shares is %d", cpuShares)
+		}
+	}
+
```
dqminh: @shishir-a412ed from the CI log, this failed because it needs to import fmt, and declare `c := m.Cgroups`

However, it's probably cleaner to check this after we set the value in `cgroups/fs/cpu.go`

shishir-a412ed:
@dqminh Thanks for that ! I have made the changes.

shishir-a412ed:
@vmarmol any updates on this ?

vmarmol:
LGTM

Argh we really need to make it easier to share code between the fs and systemd implementations. Too much duplication.

LK4D4:
@vmarmol Yeah :) We need some functions which operating on just cgroup paths.

LK4D4:
LGTM

PR #469, opened 2015-03-23
systemd: properly check DefaultDependencies is read only

Snorch:
on systemd v208 in Centos7 and Fedora20 error is not:
"org.freedesktop.DBus.Error.PropertyReadOnly"
but:
"property.org.freedesktop.DBus.Error.PropertyReadOnly"
so check failes and in Docker we get:
Docker daemon: System error: Cannot set property DefaultDependencies, or
unknown property

Fix for commit:
https://github.com/docker/libcontainer/commit/99233fde8c4f58853a474a5831ef0bcf6bf866c5

Signed-off-by: Pavel Tikhomirov <ptikhomirov@parallels.com>


crosbymichael:
LGTM

LK4D4:
LGTM

PR #472, opened 2015-03-23
path now returns the IsNotFound error

crosbymichael:
This error is not propogated up to the caller and needs to be handled at
the site where d.path() is called.


vmarmol:
LGTM

PR #474, opened 2015-03-23
Revert "cgroups: only return path when subsystem really mounted

crosbymichael:
We need to do this so that when a user specifies an option and we are unable to fulfill the request we should fail hard instead of ignoring the error and keep going.  We also need to fail hard for cgroups like `devices` because of security reasons.

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -235,6 +235,12 @@ func (raw *data) parent(subsystem string) (string, error) {
 }
 
 func (raw *data) path(subsystem string) (string, error) {
+	_, err := cgroups.FindCgroupMountpoint(subsystem)
+	// If we didn't mount the subsystem, there is no point we make the path.
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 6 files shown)


LK4D4:
@crosbymichael first commit is not signed

jfrazelle:
Probably because it was done through the UI ;)

tianon:
Shouldn't these instead only fail hard like this if both the feature was requested _and_ the `join` failed with `NotFound`?

Otherwise, any cgroup we add support for becomes required, like `blkio` did in #337 (and then revised via #437).

jfrazelle:
this is signed now :)

LK4D4:
LGTM

tianon:
c5eef90 thanks @crosbymichael :heart:

hqhq:
Hi, the problem @tianon pointed out seems still exist. This revert also revert the fix that we don't fail when some subsystems are not mounted, I'm doing this another way in https://github.com/docker/libcontainer/pull/476, so we also don't ignoring the error when a user specifies an option and we are unable to fulfill the request.

PR #475, opened 2015-03-24
remove redundant code

hqhq:
Now we return not found err in path(), so no need to check again.


crosbymichael:
LGTM

LK4D4:
LGTM

PR #476, opened 2015-03-24
don't fail when subsystem not mounted

hqhq:
We do this aim two goals:
 - don't fail when some subsystems are not mounted.
 - fail hard instead of ignoring the error when a user specifies
   an option and we are unable to fulfill the request.

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -275,6 +275,11 @@ func (raw *data) join(subsystem string) (string, error) {
 }
 
 func writeFile(dir, file, data string) error {
+	// Normally dir should not be empty, one case is that cgroup subsystem
+	// is not mounted, we will get empty dir, and we want it fail here.
+	if dir == "" {
+		return fmt.Errorf("no such directory for %s.", file)
+	}
 	return ioutil.WriteFile(filepath.Join(dir, file), []byte(data), 0700)
 }
 
--- a/cgroups/fs/cpuset.go
+++ b/cgroups/fs/cpuset.go
@@ -16,7 +16,7 @@ type CpusetGroup struct {
 
 func (s *CpusetGroup) Apply(d *data) error {
 	dir, err := d.path("cpuset")
-	if err != nil {
+	if err != nil && !cgroups.IsNotFound(err) {
 		return err
 	}
 
@@ -48,6 +48,11 @@ func (s *CpusetGroup) GetStats(path string, stats *cgroups.Stats) error {
 }
 
 func (s *CpusetGroup) ApplyDir(dir string, cgroup *configs.Cgroup, pid int) error {
+	// This might happen if we have no cpuset cgroup mounted.
+	// Just do nothing and don't fail.
+	if dir == "" {
+		return nil
+	}
 	if err := s.ensureParent(dir); err != nil {
 		return err
 	}
--- a/cgroups/fs/devices.go
+++ b/cgroups/fs/devices.go
@@ -11,6 +11,8 @@ type DevicesGroup struct {
 func (s *DevicesGroup) Apply(d *data) error {
 	dir, err := d.join("devices")
 	if err != nil {
+		// We will return error even it's `not found` error, devices
+		// cgroup is hard requirement for container's security.
 		return err
 	}
 
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -256,6 +256,11 @@ func (m *Manager) GetPaths() map[string]string {
 }
 
 func writeFile(dir, file, data string) error {
+	// Normally dir should not be empty, one case is that cgroup subsystem
+	// is not mounted, we will get empty dir, and we want it fail here.
+	if dir == "" {
+		return fmt.Errorf("no such directory for %s.", file)
+	}
 	return ioutil.WriteFile(filepath.Join(dir, file), []byte(data), 0700)
 }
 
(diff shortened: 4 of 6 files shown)


cgroups/fs/devices.go
```diff
@@ -10,7 +10,7 @@ type DevicesGroup struct {
 
 func (s *DevicesGroup) Apply(d *data) error {
 	dir, err := d.join("devices")
-	if err != nil {
+	if err != nil && !cgroups.IsNotFound(err) {
```
crosbymichael: We really cannot do this because devices is a hard requirement.  Without the devices cgroup it is a security issue because a container will have unrestricted access to device nodes.

crosbymichael:
@hqhq what cgroups does your systems not have mounted?

hqhq:
@crosbymichael some of our clients don't mount all cgroup subsystems, maybe just mount cpuset and memory cgroup. It's OK if Docker take device cgroup as hard requirement, but I think that can be done on Docker side, because that's what Docker needed, and we can show specific message, libcontainer should be more common, what do you think?

crosbymichael:
@hqhq I think cgroups are a fundamental aspect for creating containers and support should be required.  There is little overhead for mounting these and joining them as monitoring via cgroups is very important to understand the performance of your containerized apps.  I feel like containers are the sum of namespaces, cgroups, caps, and rootfs jails (chroot).  Each are equally important and should be required.  

@vmarmol how do you feel about making these mounts required or not?

vmarmol:
I see @hqhq's point about libcontainer maybe not being the place where we enforce this, but I do see the value in having a setup we know and test frequently. We already assume cgroups are mounted in various places and it'll be brittle for those that don't. I'm conflicted but I side with @crosbymichael about having a good experience for the 99% usecases.

hqhq:
@crosbymichael @vmarmol I don't think we should take all cgroup subsystems as hard requirement, devices cgroup has a good reason but not all of others. I added some check on Docker side https://github.com/docker/docker/pull/11882, and we can take devices cgroup as hard requirement in libcontaier.

hqhq:
I updated devices part, WDYT?

crosbymichael:
@hqhq ok, i think that maybe a good approach to take.

cgroups/systemd/apply_systemd.go
```diff
@@ -224,6 +224,11 @@ func (m *Manager) GetPaths() map[string]string {
 }
 
 func writeFile(dir, file, data string) error {
+	// Normally dir should not be empty, one case is that cgroup subsystem
+	// is not mounted, we will get empty dir, and we want it fail here.
+	if dir == "" {
+		return fmt.Errorf("No such directory for %s.", file)
```
vmarmol: nit: lowercase "No" for Go errors to start with lowercase.

cgroups/fs/apply_raw.go
```diff
@@ -263,6 +264,11 @@ func (raw *data) join(subsystem string) (string, error) {
 }
 
 func writeFile(dir, file, data string) error {
+	// Normally dir should not be empty, one case is that cgroup subsystem
+	// is not mounted, we will get empty dir, and we want it fail here.
+	if dir == "" {
+		return fmt.Errorf("No such directory for %s.", file)
```
vmarmol: same here.

vmarmol:
Sounds reasonable to me.

hqhq:
Updated.

mrunalp:
@hqhq Do we need any changes on the systemd side?

hqhq:
@mrunalp Yes we need indeed, updated, thanks.

hqhq:
Rebased, ping.

hqhq:
ping.

cgroups/systemd/apply_systemd.go
```diff
@@ -441,10 +448,14 @@ func joinMemory(c *configs.Cgroup, pid int) error {
...
 	path, err := getSubsystemPath(c, "cpuset")
-	if err != nil {
+	if err != nil && !cgroups.IsNotFound(err) {
 		return err
 	}
 
+	if path == "" {
+		return nil
+	}
+
```
dqminh: should this check be moved into `cpusetGroup.ApplyDir` ?

vishh:
This still doesn't satisfy the use case where no cgroups are mounted right?

hqhq:
@vishh No, for now devices cgroup is hard requirement for security.

cgroups/systemd/apply_systemd.go
hqhq: Yes, thanks, updated.

dqminh:
LGTM

cgroups/fs/cpuset.go
```diff
@@ -48,6 +48,11 @@ func (s *CpusetGroup) GetStats(path string, stats *cgroups.Stats) error {
 }
 
 func (s *CpusetGroup) ApplyDir(dir string, cgroup *configs.Cgroup, pid int) error {
```
vmarmol: Shouldn't we not even make the call to apply for this subsystem if it doesn't exist?

hqhq: I was outside, just moved in as @dqminh suggested.
If outside, we should do that on both fs and systemd side, but I don't think that's a big deal. Just want make sure which way you prefer.

calavera: Yeah, checking this outside would duplicate this logic, I like it here better too :+1:

calavera:
code LGTM.

hqhq:
ping.

dqminh:
LGTM

vmarmol:
LGTM

PR #477, opened 2015-03-24
systemd may not check DefaultDependencies is read only

cc272309126:
On system v208, when systemd unit exist, it may not check
DefaultDependencies is read only, it will return directly.
And this lead that docker container can't start correctly.
And this commit will fix this bug.

Signed-off-by: Chao Chen <chenchao@qiyi.com>

### DIFF
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -91,8 +91,14 @@ func UseSystemd() bool {
 		ddf := newProp("DefaultDependencies", false)
 		if _, err := theConn.StartTransientUnit("docker-systemd-test-default-dependencies.scope", "replace", ddf); err != nil {
 			if dbusError, ok := err.(dbus.Error); ok {
-				if strings.Contains(dbusError.Name, "org.freedesktop.DBus.Error.PropertyReadOnly") {
-					hasTransientDefaultDependencies = false
+				if strings.Contains(dbusError.Name, "org.freedesktop.systemd1.UnitExists") {
+					if err := theConn.SetUnitProperties("docker-systemd-test-default-dependencies.scope", true, ddf); err != nil {
+						if dbusError, ok := err.(dbus.Error); ok {
+							if strings.Contains(dbusError.Name, "org.freedesktop.DBus.Error.PropertyReadOnly") {
+								hasTransientDefaultDependencies = false
+							}
+						}
+					}
 				}
 			}
 		}


LK4D4:
Seems legit
ping @philips

philips:
@cc272309126 I am confused, so you hit this if "docker-systemd-test-default-dependencies.scope" already exists?

With this patch if the scope doesn't exist then you will hit a new error because DefaultDependencies won't be settable.

Can you outline how I can reproduce this a bit more?

cc272309126:
@philips I reproduce it by next steps, 
env
kernel：3.10.0-123.el7.x86_64, distribution：CentOS Linux release 7.0.1406 (Core)
systemd：systemd-208-11.el7_0.6.x86_64
steps：
1.use /usr/bin/docker -D -d to start docker
2.run the command "docker run -i -t image_id /bin/bash" and "exit“
3.check the dir /run/systemd/system/, we will found "docker-systemd-test-default-dependencies.scope" already exists. 
4.after steps 3, run "docker run -i -t image_id /bin/bash" again. it will  lead that docker container can't start correctly.
5.the reason may be that systemd will return an exist error if the unit is exist。and we may can upgrade systemd to avoid this issue.

philips:
@cc272309126 OK, I will attempt to reproduce and fix.

cc272309126:
@philips thank you very much

philips:
My plan of attack: 

- Introduce some entropy to the unit name (probably process ID) in case the user is running two dockers
- attempt to use Kill() on the scope or make the transient unit a service instead of scope

philips:
@cc272309126 Would you mind testing out this branch? https://github.com/philips/libcontainer/commits/stop-test-unit

cc272309126:
@philips ok, I will try it

cc272309126:
@philips it's ok for me. thank you for fix it.

cc272309126:
@philips I test more detaily. I have a doubt. 
 I think  this branch 's "testScopeExists" flag will be true always. because the libcontainer-%i-systemd-test-default-dependencies.scope may not be exists always. if this happend. the “DefaultDependencies” property may not be checked. am I right?

philips:
@cc272309126 I don't think this is correct. If it successfully stops the unit it will get "org.freedesktop.systemd1.NoSuchUnit", at which time testScopeExists gets set to false and that should skip over the short circuit.

cc272309126:
@philips ok, I got it. this patch LGTM. thank you

LK4D4:
@philips What we should do? Seems like RHEL7 is affected.

philips:
@LK4D4 I will send a PR for my patch, I suppose.

philips:
Posted #489

LK4D4:
@philips Thanks!

tiborvass:
@LK4D4 shouldn't this be closed if #489 replaces this?

crosbymichael:
@tiborvass yes.

@cc272309126 thanks for your help with this bug

PR #479, opened 2015-03-24
WIP: Add Checkpoint and Restore support to libcontainer

crosbymichael:
This is a work in progress PR so that we can move the discussion around this work into a PR to help move things along and get other people involved.

We are trying to get `nsinit` work as an easy way to test but the first class support should be for using it from a daemon such as docker.  

Any more work can be done on the `criu` branch of this repo via pull requests.


configs/config.go
```diff
@@ -96,6 +96,9 @@ type Config struct {
 	// ReadonlyPaths specifies paths within the container's rootfs to remount as read-only
 	// so that these files prevent any writes.
 	ReadonlyPaths []string `json:"readonly_paths"`
+
+	// Container's standard descriptors (std{in,out,err}), needed for checkpoint and restore
+	StdFds [3]string `json:"ext_pipes,omitempty"`
```
mrunalp: Minor nit: Could the json property be called something else?

LK4D4:
Supercool.
Gofmt failed.

jfrazelle:
so cool so cool so cool, can't contain myself

On Tue, Mar 24, 2015 at 1:49 PM, Alexander Morozov <notifications@github.com

container_linux.go
```diff
@@ -249,6 +257,113 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+func (c *linuxContainer) Checkpoint() error {
+	c.m.Lock()
+	defer c.m.Unlock()
+	dir := filepath.Join(c.root, "checkpoint")
+	// Since a container can be C/R'ed multiple times,
+	// the checkpoint directory may already exist.
+	if err := os.Mkdir(dir, 0655); err != nil && !os.IsExist(err) {
+		return err
+	}
+	args := []string{
```
mrunalp: I know we have hashed this before :) Should we check for the presence of a minimum version of criu that supports all the flags or maybe just document it, so distributions know what version they need?

vishh:
:+1:

container_linux.go
SaiedKazemi: Most of the flags have been there for many years and we haven't introduced new flags in a long time.  But if someone is running an old version of CRIU that doesn't support a specific flag, CRIU will print an error message and exit -- no damage done.  That said, we can certainly add code to do version check every time but I really think it'd be an overkill :)

mrunalp: @SaiedKazemi That is fair. I think if we just document a minimum version, that will be helpful :)

SaiedKazemi: Sounds good.

configs/config.go
SaiedKazemi: @mrunalp 
Sure, any suggestions?  The name ext_pipes was chosen because the standard file descriptors are actually external pipes but I am open to better suggestions.

mrunalp: @SaiedKazemi I was thinking that the name of the exported field and the json tag be similar like elsewhere in the code. If the name is StdFds, then maybe std_fds for the json tag.

hqhq:
Looking forward... :+1:

container_linux.go
vishh: +1 for validating compatibility.

container_linux.go
```diff
@@ -249,6 +257,124 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
 	return notifyOnOOM(c.cgroupManager.GetPaths())
 }
 
+// XXX debug support, remove when debugging done.
```
vishh: Can we get rid of all the debug code?

configs/config.go
LK4D4: Yes, +1. or rename StdFds to ExtPipes :)

container_linux.go
```diff
@@ -213,6 +215,12 @@ func newPipe() (parent *os.File, child *os.File, err error) {
 func (c *linuxContainer) Destroy() error {
 	c.m.Lock()
 	defer c.m.Unlock()
+	// Since the state.json and CRIU image files are in the c.root
+	// directory, we should not remove it after checkpoint.  Also,
+	// when     CRIU exits after restore, we should not kill the processes.
+	if _, err := os.Stat(filepath.Join(c.root, "checkpoint")); err == nil {
```
LK4D4: I have question about `c.root`, because now we supposing that all in that root is not critical and can be removed on reboot or containers cleaning. But c/r data is pretty critical, so it should be persisted, probably makes sense to add `CRIURoot`?

LK4D4: But maybe makes sense not supposing that root shouldn't persist.

container_linux.go
```diff
@@ -249,6 +257,124 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+		"-t", strconv.Itoa(c.initProcess.pid()),
+	}
+	for _, m := range c.config.Mounts {
+		if m.Device == "bind" {
+			args = append(args,
+				"--ext-mount-map", fmt.Sprintf("%s:%s", m.Destination, m.Destination))
+		}
+	}
+	addArgsFromEnv("CRIU_C", &args) // XXX debug
+	if err := exec.Command(c.criuPath, args...).Run(); err != nil {
```
LK4D4: Isn't makes sense to use `CombinedOutput` and print it in case of error?

crosbymichael:
Just for an FYI, `nsinit` is just for testing, the use case we should optimize for is the daemon(docker) use cases.  We should have the ability to make checkpoint/restore options configurable.

container_linux.go
vishh: +1

container_linux.go
```diff
@@ -249,6 +257,124 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+	for i := 0; i < 3; i++ {
+		if s := c.config.StdFds[i]; strings.Contains(s, "pipe:") {
+			args = append(args, "--inherit-fd", fmt.Sprintf("fd[%d]:%s", i, s))
+		}
+	}
+	addArgsFromEnv("CRIU_R", &args) // XXX debug
+
+	// XXX This doesn't really belong here as our caller should have
+	//     already set up root (including devices) and mounted it.
+	/*
```
vishh: Delete dead code?

configs/config.go
SaiedKazemi: Sounds good.  Will do.

container_linux.go
SaiedKazemi: @LK4D4 @crosbymichael 
For Docker containers, I use "criu_img" subdirectory in "/var/lib/containers/<ID>" which is outside the container's root.  I think @crosbymichael used the "checkpoint" subdirectory in "/nsinit" of the container because nsinit is really a test program.  In general, I agree with you that we do not want to save the files that CRIU generates (images, logs, and pid) inside container's root.

container_linux.go
SaiedKazemi: @vishh 
Since we're still debugging, it makes a lot easier to add additional command line options when invoking CRIU through the environment.  Otherwise, we need to edit and recompile source for every little change in the options.  I will remove it soon.

container_linux.go
SaiedKazemi: @LK4D4 @vishh 
CRIU logs everything in the log files and generally doesn't output.

container_linux.go
SaiedKazemi: @vishh 
Will do.

boucher:
I'm interested in a slightly different use case than possible with the way this PR would work. Specifically, I want to be able to checkpoint a container without stopping it, and then later be able to either "roll back" the container to that point in time or alternatively spin up a new container from that checkpoint. Is there any interest in the idea of allowing you to specify an explicit path to checkpoint to and restore from (and, additionally, flags like --leave-running).

boucher:
I'm trying to get this pull request running and having some trouble. What I've tried so far is to build libcontainer (with it's Dockerfile, and the addition of criu to that Dockerfile), and run the following test:

    docker run -d --privileged -w /busybox dockercore/libcontainer nsinit exec -- sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 3; done'

followed by `nsinit checkpoint` from within the resulting container. That fails, with this output log:

https://gist.github.com/boucher/d2539bcaa1311c5d5f5c

Any suggestions? Or, is there a better way to test this?

boucher:
Just wanted to follow up in this thread that I was able to get this PR running outside of Docker, directly on a linux host, but haven't yet been able to get it working inside of Docker.

container_linux.go
```diff
@@ -252,6 +260,124 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+	cmd.Stdin = process.Stdin
+	cmd.Stdout = process.Stdout
+	cmd.Stderr = process.Stderr
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	// cmd.Wait() waits cmd.goroutines which are used for proxying file descriptors.
+	// Here we want to wait only the CRIU process.
+	st, err := cmd.Process.Wait()
```
boucher: I'm working on getting this branch up and running in Docker, but I'm a bit unsure about what to do here. Because we're not passing the -d flag to criu, this Wait will actually just wait forever -- that's fine in the sort of direct nsinit resume an interactive container case (which I had used to test this branch), but it doesn't work when dealing with the Docker daemon. Should this be a configurable flag we're passing in, or should the wait be handled at the nsinit level and not here?

avagin: Yes, here is a mistake. I've fixed it in https://github.com/docker/libcontainer/pull/486

container_linux.go
```diff
@@ -252,6 +255,336 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+
+func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Process, imagePath string) error {
+	notify := resp.GetNotify()
+	if notify == nil {
+		return fmt.Errorf("invalid response: %s", resp.String())
+	}
+
+	switch {
+	case notify.GetScript() == "post-dump":
+		f, err := os.Create(filepath.Join(c.root, "checkpoint"))
```
boucher: Is the purpose of this file now just to be a flag that things are checkpointed? (Previously, this is where things were being stored if no directory was specified)

container_linux.go
```diff
@@ -252,6 +255,336 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+		if err != nil {
+			return err
+		}
+		f.Close()
+		break
+
+	case notify.GetScript() == "post-restore":
+		// In many case, restore from the images can be done only once.
+		// If we want to create snapshots, we need to snapshot the file system.
+		os.RemoveAll(imagePath)
```
boucher: It seems shortsighted to assume that the image should be deleted after a restore. In particular, if I'm snapshotting the state of the filesystem and storing filesystem snapshots as well, I should be able to restore multiple times from that. (I'm working on fully checkpointing a docker container, fs and process state entirely, so that i can essentially run copies of it, which would involve multiple restores)

container_linux.go
avagin: @boucher Yes, it's a flag. It's used in currentStatus() to determine the Checkpointed state.

boucher:
I'm getting this error on some restore attempts when running this branch:

    Error (cr-service.c:58): Failed unpacking request: Success
    Error (cr-service.c:694): Can't recv request: Success
    data too short after length-prefix of 1217

I added a log in criuSwrk to print all the opts right before actually sending them, and other than the actual directory paths I seem to have two identical set of options that work and don't work (though the underlying container isn't the same, but the error message seems to imply a communication failure not an actual restore failure). I'm running in two different VMs so I'm going to try and narrow down the differences some more to figure out what's going on but I figured I'd post this here to see if there were any ideas.

avagin:
@boucher I think we use too small buffer in CRIU. Probably we need to allocate it dinamically. Could you try out this patch for criu?
```
diff --git a/include/cr-service-const.h b/include/cr-service-const.h
index 668882b..66a059f 100644
--- a/include/cr-service-const.h
+++ b/include/cr-service-const.h
@@ -1,7 +1,7 @@
 #ifndef __CR_SERVICE_CONST_H__
 #define __CR_SERVICE_CONST_H__
 
-#define CR_MAX_MSG_SIZE 1024
+#define CR_MAX_MSG_SIZE 16384
 #define CR_DEFAULT_SERVICE_ADDRESS "/var/run/criu_service.socket"
 
 #endif /* __CR_SERVICE_CONST_H__ */
```

container_linux.go
SaiedKazemi: @boucher 
I agree that we should not remove the images directory.  In addition to the use case you've mentioned, these images help in forensic debugging of the container processes.

avagin:
The last issue is a criu bug: http://lists.openvz.org/pipermail/criu/2015-April/019928.html

avagin:
@crosbymichael @SaiedKazemi @LK4D4 @mrunalp @vmarmol I think I've fixed all known issue. Maybe it's time to merge this branch?

SaiedKazemi:
@avagin @crosbymichael @LK4D4 @vmarmol 
Thanks Andrew.  I agree it's time to merge into master.

crosbymichael:
Ok, so I should be able to build this branch, and everything with work flawlessly the first time?

avagin:
@crosbymichael @SaiedKazemi I forget about network. It's practically unsupported now.

boucher:
I am hoping to put together a pull request tonight or tomorrow with my changes to support passing more flags to criu, but I could do that against master if things get merged first.

container_linux.go
```diff
@@ -252,6 +255,338 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
...
+	}
+	defer workDir.Close()
+
+	imageDir, err := os.Open(imagePath)
+	if err != nil {
+		return err
+	}
+	defer imageDir.Close()
+
+	root := filepath.Join(c.root, "criu-root")
```
boucher: Why do we make this root directory here during restore, as opposed to just using the same root checkpoint used (c.config.Rootfs)?

avagin: @boucher CRIU has a few requirements for a root directory:
* it must be a mount point
* its parent must not be overmounted

We create a temporary directory only to satisfy these requirements.

boucher:
What (if anything) needs to be done before this could be merged?

crosbymichael:
Does it work now?  When I go to test this it should be effortless and easy and it should work the first time ;)

Does this required a special version of criu now that I have to build from source?  If so, please give me a commit to build from or i'll just build HEAD.

avagin:
@crosbymichael Yes, it works except unknown bugs;). Feel free to report them.

We've released criu 1.5.2 and it should be used.
http://lists.openvz.org/pipermail/criu/2015-April/020172.html
https://github.com/xemul/criu/commit/4e3a14ce6120df890e281bc038a0579aa6a011c1

Sirupsen:
This is really exciting work! Is there any update on the status of this PR?

crosbymichael:
I'm getting the error:

```bash
nsinit checkpoint --image-path /tmp/checkpoint
criu failed: type DUMP errno 0
```

crosbymichael:
criu is compiled via the v1.5.2 tag

boucher:
Can you send the dump log?

On Thursday, May 21, 2015, Michael Crosby <notifications@github.com> wrote:

> criu is compiled via the v1.5.2 tag
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/libcontainer/pull/479#issuecomment-104384896>.
>


-- 
Sent from Gmail Mobile

crosbymichael:
@boucher i don't have the dump.log in the image dir? is it located somewhere else?

avagin:
@crosbymichael Could you attache a strace log?
strace -fo strace.log -s 256 nsinit checkpoint --image-path /tmp/checkpoint

crosbymichael:
@avagin https://gist.github.com/f8f2a349ecd9ad6f62cc

avagin:
@crosbymichael this log doesn't contain "criu failed:...". dump.log is saved in /var/run/nsinit/nsinit/criu.work/

'''
write(1023, "(00.235986) Dumping finished successfully\n", 42) = 42
'''

crosbymichael:
ok, it was a setuid issue with my binary

crosbymichael:
LGTM

crosbymichael:
thanks @avagin @boucher @SaiedKazemi

PR #480, opened 2015-03-25
mount: Add a flag to bind devices when user namespaces are enabled.

mrunalp:
Minor refactor. 


rootfs_linux.go
```diff
@@ -218,6 +213,13 @@ func createDeviceNode(rootfs string, node *configs.Device) error {
 			f.Close()
 		}
 		return syscall.Mount(node.Path, dest, "bind", syscall.MS_BIND, "")
+	} else {
```
crosbymichael: We don't need this else if we have a return right above it correct?

mrunalp: Done.

crosbymichael:
LGTM

PR #484, opened 2015-03-25
Process capabilities

mrunalp:
This allows a process to override the capabilities specified in the container config. This could allow one to run more/less privileged processes in a container. For e.g. to examine logs or modify iptables.

### DIFF
--- a/init_linux.go
+++ b/init_linux.go
@@ -40,13 +40,14 @@ type network struct {
 
 // initConfig is used for transferring parameters from Exec() to Init()
 type initConfig struct {
-	Args     []string        `json:"args"`
-	Env      []string        `json:"env"`
-	Cwd      string          `json:"cwd"`
-	User     string          `json:"user"`
-	Config   *configs.Config `json:"config"`
-	Console  string          `json:"console"`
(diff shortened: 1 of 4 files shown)


init_linux.go
```diff
@@ -99,7 +100,12 @@ func finalizeNamespace(config *initConfig) error {
 	if err := utils.CloseExecFrom(3); err != nil {
 		return err
 	}
-	w, err := newCapWhitelist(config.Config.Capabilities)
+
+	capabilities := config.Config.Capabilities
```
LK4D4: Looks super-weird :) But seems right.

mrunalp: Ha, ha yes, when I was looking at it before sending PR, I was sure I had done something wrong.

LK4D4:
LGTM

crosbymichael:
LGTM

crosbymichael:
@mrunalp do you think we should just remove the caps on the config or keep both as process can override?

mrunalp:
@crosbymichael I think we should keep both, as most times processes could be started with the same capabilities as specified in the container config and can override them only if necessary.

PR #485, opened 2015-03-26
ct: initialize StdFds for restored containers

avagin:
Cc: @SaiedKazemi


avagin:
BTW: http://lists.openvz.org/pipermail/criu/2015-March/019400.html

crosbymichael:
Shouldn't stdfds be part of `Process` and not config?

avagin:
We are going to use another version of this code https://github.com/docker/libcontainer/pull/486

PR #487, opened 2015-03-27
Use syscall.Kill instead of p.cmd.Process.Kill

LK4D4:
We need this to unmask syscall.ESRCH error, which handled in docker and
can be handled by other clients.

Closes #457


LK4D4:
Also I encountered this problem myself with running long-running containers and trying to remove them in parallel.

crosbymichael:
LGTM

PR #492, opened 2015-03-31
cgroups: add support of devices deny for another use of cgroup devices

Mashimiao:

the original code usee cgroup devices, like:
group        whitelist entries                        denied devices
A                "c x:x rwm", "b x:* rwm"                 all the rest
I add another use method, you can pass parameters, like:
group        whitelist entries                        denied devices
A                all                                            "c x:x rwm", "b x:* rwm"

### DIFF
--- a/cgroups/fs/devices.go
+++ b/cgroups/fs/devices.go
@@ -32,6 +32,17 @@ func (s *DevicesGroup) Set(path string, cgroup *configs.Cgroup) error {
 				return err
 			}
 		}
+		return nil
+	}
+
+	if err := writeFile(path, "devices.allow", "a"); err != nil {
+		return err
+	}
+
+	for _, dev := range cgroup.DeniedDevices {
+		if err := writeFile(path, "devices.deny", dev.CgroupString()); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/cgroups/fs/devices_test.go
+++ b/cgroups/fs/devices_test.go
@@ -17,7 +17,18 @@ var (
 			FileMode:    0666,
 		},
 	}
-	allowedList = "c 1:5 rwm"
+	allowedList   = "c 1:5 rwm"
+	deniedDevices = []*configs.Device{
+		{
+			Path:        "/dev/null",
+			Type:        'c',
+			Major:       1,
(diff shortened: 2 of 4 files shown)


hqhq:
looks good.

Mashimiao:
ping @LK4D4 , @mrunalp

dqminh:
LGTM

Mashimiao:
cc @docker/distribution-maintainers

Mashimiao:
ping @LK4D4

cgroups/fs/devices.go
```diff
@@ -32,6 +32,16 @@ func (s *DevicesGroup) Set(path string, cgroup *configs.Cgroup) error {
 				return err
 			}
 		}
+	} else {
```
LK4D4: You should `return nil` in `if` block and not use `else`. Same for systemd.

LK4D4:
@Mashimiao Sorry for long response. I missed ping in flood of mail.
PR is supercool. Could you write integration test? Let me know if you have no time - we can write test later.
Also need rebase.

cgroups/fs/devices.go
Mashimiao: @LK4D4 Could you give more explanation about why should not use return nil and else?

Mashimiao:
@LK4D4 I rebaseed first, will fix code after I know why. 
I think I need more time to understand how integration works before writing integration test. 
I think writing test later is a better choice

cgroups/fs/devices.go
hqhq: I think what @LK4D4 means is you should write code like this:
```
if xxx {
    xxx
    return nil
}

your code here

return nil
```

LK4D4: Yup, thanks @hqhq!

Mashimiao: OK, Fixed.

LK4D4:
Ok for me.
ping @crosbymichael

Mashimiao:
ping @crosbymichael

dqminh:
LGTM too

crosbymichael:
LGTM

PR #494, opened 2015-03-31
cgroups: reurn error when passing invalid argument to freezer

Mashimiao:

### DIFF
--- a/cgroups/fs/freezer.go
+++ b/cgroups/fs/freezer.go
@@ -1,6 +1,7 @@
 package fs
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


LK4D4:
@Mashimiao Tests fail :)

cgroups/fs/freezer.go
```diff
@@ -41,6 +42,8 @@ func (s *FreezerGroup) Set(path string, cgroup *configs.Cgroup) error {
 			}
 			time.Sleep(1 * time.Millisecond)
 		}
+	default:
+		fmt.Errorf("Invalid argument %s to freezer.state", cgroup.Freezer)
 	}
 
 	return nil
```
LK4D4: You can remove it now. Because it's unreachable code.

Mashimiao: Hi LK4D4, Thanks for reviewing.
It seems it is.
But what I mainly want to do is to return error clearly when people pass  invalid argument accidentally.
I added new line testing configs.Undefined. How do you think about it?

LK4D4:
LGTM
Also you have misspell in commit message `reurn`. You can fix if you want.

mrunalp:
LGTM

PR #495, opened 2015-03-31
Add support for Premount and Postmount commands.

rhatdan:
We want to allow docker to mount tmpfs directories over existing directories
in the image.  We will use this patch to pass commands from docker to
libcontainer.  The first command we will use is the tar command to gather
all of the contents of the destination directory before mounting, then after
we mount the post mount command will untar the content.

### DIFF
--- a/configs/mount.go
+++ b/configs/mount.go
@@ -18,4 +18,17 @@ type Mount struct {
 
 	// Relabel source if set, "z" indicates shared, "Z" indicates unshared.
 	Relabel string `json:"relabel"`
+
+	// Optional Command to be run before Source is mounted.
+	PremountCmds []Command `json:"premount_cmds"`
+
+	// Optional Command to be run after Source is mounted.
+	PostmountCmds []Command `json:"postmount_cmds"`
+}
+
+type Command struct {
+	Path string   `json:"path"`
+	Args []string `json:"args"`
+	Env  []string `json:"env"`
+	Dir  string   `json:"dir"`
 }
--- a/rootfs_linux.go
+++ b/rootfs_linux.go
@@ -80,7 +80,7 @@ func mountCmd(cmd configs.Command) error {
 	command.Env = cmd.Env
 	command.Dir = cmd.Dir
 	if out, err := command.CombinedOutput(); err != nil {
-		return fmt.Errorf("%s failed: %s: %v", cmd, string(out), err)
+		return fmt.Errorf("%#v failed: %s: %v", cmd, string(out), err)
 	}
 
 	return nil
--- a/rootfs_linux.go
+++ b/rootfs_linux.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"syscall"
@@ -24,8 +25,19 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 		return newSystemError(err)
 	}
 	for _, m := range config.Mounts {
+		for _, precmd := range m.PremountCmds {
+			if err := mountCmd(precmd); err != nil {
+				return newSystemError(err)
+			}
+		}
 		if err := mountToRootfs(m, config.Rootfs, config.MountLabel); err != nil {
 			return newSystemError(err)
+		}
+
+		for _, postcmd := range m.PostmountCmds {
+			if err := mountCmd(postcmd); err != nil {
+				return newSystemError(err)
+			}
 		}
 	}
 	if err := createDevices(config); err != nil {
@@ -59,6 +71,18 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 		}
 	}
 	syscall.Umask(0022)
+	return nil
+}
+
+func mountCmd(cmd configs.Command) error {
+
+	command := exec.Command(cmd.Path, cmd.Args[:]...)
+	command.Env = cmd.Env
+	command.Dir = cmd.Dir
+	if out, err := command.CombinedOutput(); err != nil {
+		return fmt.Errorf("%s failed: %s: %v", cmd, string(out), err)
+	}
+
 	return nil
 }
 
--- a/integration/exec_test.go
+++ b/integration/exec_test.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/docker/libcontainer"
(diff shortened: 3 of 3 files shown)


rootfs_linux.go
```diff
@@ -62,6 +69,21 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 	return nil
 }
 
+func mountCmd(cmd string) error {
+	if cmd != "" {
+		args := strings.Split(cmd, " ")
+		mCmd := exec.Command(args[0], args[1:]...)
```
mrunalp: I guess could just use CombinedOutput here instead of breaking into steps.

rootfs_linux.go
```diff
@@ -62,6 +69,21 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
...
+		if err := mCmd.Start(); err != nil {
```
shishir-a412ed: @rhatdan we can use mCmd.Run() here instead of {Start, Wait} commands.
http://golang.org/pkg/os/exec/#Cmd.Run

rootfs_linux.go
rhatdan: Not sure what you mean?

Will
+ mCmd := exec.Command(args...)
work?

rhatdan: # github.com/docker/libcontainer
vendor/src/github.com/docker/libcontainer/rootfs_linux.go:74: not enough arguments in call to exec.Command

shishir-a412ed: @mrunalp do you mean something like:
`return exec.Command(args[0], args[1:]...).Run()`

LK4D4: `return exec.Command(args[0], args[1:]...).CombinedOutput()` it makes sense to check output of command on error.

mrunalp: If you want to capture output in case of failure and return it back:
```go
output, err := exec.Command(args[0], args[1:]...).CombinedOutput()
if err != nil {
       return fmt.Errorf("Mounting failed: %v, err, output)
}
```
If not, then Run is fine like you have it now.

rootfs_linux.go
```diff
@@ -62,6 +69,17 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 	return nil
 }
 
+func mountCmd(cmd string) error {
+	if cmd != "" {
+		args := strings.Split(cmd, " ")
+		out, err := exec.Command(args[0], args[1:]...).CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("%s failed: %s", cmd, string(out))
```
mrunalp: Maybe also return err embedded inside so we can catch issues with exec'ing the command itself?

configs/mount.go
```diff
@@ -18,4 +18,10 @@ type Mount struct {
 
 	// Relabel source if set, "z" indicates shared, "Z" indicates unshared.
 	Relabel string `json:"relabel"`
+
+	// Optional Command to be run before Source is mounted.
+	PremountCmd string `json:"premountcmd"`
```
dqminh: i think that this's better as a list of commands, as user should be able to perform several commands together.

rootfs_linux.go
```diff
@@ -24,9 +25,15 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 		return newSystemError(err)
 	}
 	for _, m := range config.Mounts {
+		if err := mountCmd(m.PremountCmd); err != nil {
```
dqminh: Do you think it's better to setup some special env variables like rootfs path ( similar to how lxc does it ) so commands can be self-aware of the container rootfs, or is it enough for the caller ( i.e., docker ) to pass them in as arguments ?

configs/mount.go
rhatdan: That would allow us to remove the tar ball, in the PostmountCmd. which would be nice.

rootfs_linux.go
rhatdan: Ok

mrunalp:
@avagin Yes, good point. +1.

dqminh:
@rhatdan @mrunalp i had some free time today, so I added an integration test for this at https://github.com/rhatdan/libcontainer/pull/1 Hope it helps.

rhatdan:
@dqminh Added.

configs/mount.go
```diff
@@ -18,4 +18,10 @@ type Mount struct {
 
 	// Relabel source if set, "z" indicates shared, "Z" indicates unshared.
 	Relabel string `json:"relabel"`
+
+	// Optional Command to be run before Source is mounted.
+	PremountCmds [][]string `json:"premountcmd"`
+
+	// Optional Command to be run after Source is mounted.
+	PostmountCmds [][]string `json:"postmountcmd"`
```
dqminh: nit: I think the json labels should be `premount_cmds` and `postmount_cmds` to reflect their attributes.

rootfs_linux.go
```diff
@@ -62,6 +74,13 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 	return nil
 }
 
+func mountCmd(cmd []string) error {
```
dqminh: I think we should check for length of `cmd` here, it should not panic if cmd is empty

dqminh:
LGTM

configs/mount.go
```diff
@@ -18,4 +18,10 @@ type Mount struct {
 
 	// Relabel source if set, "z" indicates shared, "Z" indicates unshared.
 	Relabel string `json:"relabel"`
+
+	// Optional Command to be run before Source is mounted.
+	PremountCmds [][]string `json:"premount_cmds"`
```
LK4D4: Maybe we can use some structure which looks like `exec.Cmd`? I think Env and Dir parameters can be useful.

rhatdan:
@crosbymichael @mrunalp What do you think, would you want to pass 

https://golang.org/pkg/os/exec/#Cmd across?

avagin:
You don't ask me, but I think it is not necessary now and can be extended in a future if it will be required. And how are you going to save exec.Cmd in json format as a part of a config file.

LK4D4:
@avagin I proposing to use struct instead of `[]string`, not exactly `exec.Cmd`. Changing this later will break API.

avagin:
Sorry, I've misunderstood you. Now this idea looks reasonable to me.

rhatdan:
@LK4D4 Get @mrunalp and or @crosbymichael to agree with you and I will have @shishir-a412ed work on it...  Do not want to waste more time on this patch if it still will not get merged.

shishir-a412ed:
@LK4D4 whats your thought behind using structs instead of [][]string.
With current implementation you can inject as many commands as you want, in one call to libcontainer and libcontainer will handle all of those.

Eventually libcontainer will generate that struct using `exec.Cmd` before executing it.
I think this is a clean solution, unless you have something specific in mind.

rhatdan:
@LK4D4 @shishir-a412ed points out that we want docker to pass two commands in the postcmd section.
premount tar cf /run/foobar.tar /run

Postmount tar xf /run/foobar.tar
Postmount rm -f /run/foobar.tar

I guess we could setup a list of cmd structs.

mrunalp:
I think having a list of Command structs is reasonable. 
The struct could be a simplified version of standard library Cmd. @LK4D4 WDYT?
```go
type Command struct {
    Path string
    Args []string
    Env  []string
}

**Comment by dqminh** [comment]
@mrunalp @LK4D4 maybe also add Dir to specify the working directory ? So
```go
type Command struct {
    Path string      `json:"path"`
    Args []string    `json:"args"`
    Env  []string    `json:"env"`
    Dir  string    `json:"dir"`
}
```

LK4D4:
Yup, I like version of @dqminh . Not sure if `Dir` can do some harm, because you can do `cd` anyway with `sh` as `Path`.

mrunalp:
Works for me, too.

rhatdan:
@crosbymichael @mrunalp @LK4D4 I think Shishir has hit all of your points?

LK4D4:
LGTM

rootfs_linux.go
```diff
@@ -62,6 +74,18 @@ func setupRootfs(config *configs.Config, console *linuxConsole) (err error) {
 	return nil
 }
 
+func mountCmd(cmd configs.Command) error {
+
+	command := exec.Command(cmd.Path, cmd.Args[:]...)
+	command.Env = cmd.Env
+	command.Dir = cmd.Dir
+	if out, err := command.CombinedOutput(); err != nil {
+		return fmt.Errorf("%s failed: %s: %v", cmd, string(out), err)
```
dqminh: nit: i like `%#v` for formating cmd because it's an object

dqminh:
LGTM other than one small nit

mrunalp:
LGTM

rhatdan:
WooHoo, when will libcontainer be merged into docker?

LK4D4:
@rhatdan Whenever you need it for PR.

rhatdan:
Already do

https://github.com/docker/docker/pull/8478

LK4D4:
Ah, wait! I wanted to use version tags for libcontainer! :)
But we'll update it in a day or so.

PR #496, opened 2015-03-31
add option to pass additional fds to container process

Mic92:
This can be usefull to implement socket activated containers for example.

### DIFF
--- a/container_linux.go
+++ b/container_linux.go
@@ -139,7 +139,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
-	cmd.ExtraFiles = []*os.File{childPipe}
+	cmd.ExtraFiles = append([]*os.File{childPipe}, p.ExtraFiles...)
 	// NOTE: when running a container with no PID namespace and the parent process spawning the container is
 	// PID1 the pdeathsig is being delivered to the container's init process by the kernel for some reason
 	// even with the parent still running.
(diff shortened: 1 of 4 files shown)


mrunalp:
@Mic92 Build failed on DCO validation https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

Mic92:
I always forgot that.

container_linux.go
```diff
@@ -135,6 +135,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	cmd.Stdin = p.Stdin
 	cmd.Stdout = p.Stdout
 	cmd.Stderr = p.Stderr
+	cmd.ExtraFiles = p.ExtraFiles
```
crosbymichael: This is not going to work because of line 143 just below it

Mic92: Is the file descriptor `childPipe` leaked to the init process?

crosbymichael: The correct code would look like:

`cmd.ExtraFiles = append(p.ExtraFiles, childPipe)`

Mic92: A nil check is needed i think:

```
 extraFiles = []*os.File{childPipe}
 if p.ExtraFiles != nil {
         append(extraFiles, p.ExtraFiles)
 }
 cmd.ExtraFiles = extraFiles
```
but I need to figure out, if the passed files gets the right inode numbers (from 3 to n)

avagin:
Could you add a test for this into integration/?

container_linux.go
crosbymichael: Ah, good point, we want the `childPipe` to get fd3 so it needs to go first.  In go you can do this:

`cmd.ExtraFiles = append([]*os.File{childPipe}, p.ExtraFiles...)` to do it all in one line

Mic92:
* Passing file descriptors works now
* Added tests

container_linux.go
```diff
@@ -139,10 +139,11 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
-	cmd.ExtraFiles = []*os.File{childPipe}
 	// NOTE: when running a container with no PID namespace and the parent process spawning the container is
 	// PID1 the pdeathsig is being delivered to the container's init process by the kernel for some reason
 	// even with the parent still running.
+	cmd.ExtraFiles = append([]*os.File{childPipe}, p.ExtraFiles...)
+	cmd.SysProcAttr.Pdeathsig = syscall.SIGKILL
```
crosbymichael: I think this is a failed rebase and needs to be removed.  Also the comment above cmd.ExtraFiles needs to be on the if block for the ParentDeathSignal.

init_linux.go
```diff
@@ -95,10 +96,22 @@ func populateProcessEnvironment(env []string) error {
...
+			return err
+		}
+	}
+
+	// Ensure that all unwanted fds we may have accidentally
 	// inherited are marked close-on-exec so they stay out of the
 	// container
-	if err := utils.CloseExecFrom(3); err != nil {
+
+	if err := utils.CloseExecFrom(config.PassedFilesCount + 3); err != nil {
```
crosbymichael: Nice find!

Mic92:
Fix Rebase.

PR #500, opened 2015-04-02
add Set support for systemd based cgroup

hqhq:

### DIFF
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -347,7 +347,17 @@ func (m *Manager) GetStats() (*cgroups.Stats, error) {
 }
 
 func (m *Manager) Set(container *configs.Config) error {
-	panic("not implemented")
+	for name, path := range m.Paths {
+		sys, ok := subsystems[name]
+		if !ok || !cgroups.PathExists(path) {
+			continue
+		}
+		if err := sys.Set(path, container.Cgroups); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func getUnitName(c *configs.Cgroup) string {


Mashimiao:
As cgroup/systemd is a module using systemd to manage cgroup.
I think the Set() should not work in the same way as Set() in cgroup/fs/*.
We should let it work by systemd's methods.

Actually, I wonder why do we need systemd module?

hqhq:
Yeah, technically systemd should have it's own way, but as lack of a lot of functions, it's difficult even not possible to make systemd do all the job in systemd's way right now, so we have a lot of work around, I'm afraid `Set` would be one of them, just like `GetStats`.

Why do we need systemd module? Good question, @crosbymichael @LK4D4 I'm also looking for the answer.

mrunalp:
@hqhq Sooner or later cgroups are going to be more tightly integrated with systemd and hence it makes sense to support it as more distributions are switching to systemd. (Just my take).

Mic92:
@mrunalp what kind of integration could see for example? I think cgroups will never depend on systemd. And when using systemd, I currently do not see advantages in supporting scope units, if it is just used for cgroup accounting. It might be required in the past, when systemd did not allow to partition the its own service cgroup further, but since [systemd v218](http://lists.freedesktop.org/archives/systemd-devel/2014-December/026189.html), systemd adds the Delegate property for unit files to allow this even for unprivileged users.

hqhq:
ping.

crosbymichael:
LGTM

LK4D4:
Yeah, for now we can stick to fs implementation for systemd and switch to D-Bus later.

LK4D4:
LGTM

PR #503, opened 2015-04-02
Add ChargeProcess to Container API.

vmarmol:
This will allow charging of an existing processes' resources to a particular container. It does NOT enter the namespaces (since it can't really). Useful for resource-only containers as well as charging system daemons to containers when they are doing work on behalf of the container.

/cc @crosbymichael @vishh @avagin @rjnagal

### DIFF
--- a/cgroups/cgroups.go
+++ b/cgroups/cgroups.go
@@ -34,6 +34,9 @@ type Manager interface {
 
 	// Set the cgroup as configured.
 	Set(container *configs.Config) error
+
+	// Enters the specified process into the cgroup set.
+	AddProcess(pid int) error
 }
 
 type NotFoundError struct {
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -161,6 +161,20 @@ func (m *Manager) Set(container *configs.Config) error {
 	return nil
 }
 
+func (m *Manager) AddProcess(pid int) error {
+	for name, path := range m.Paths {
+		_, ok := subsystems[name]
+		if !ok || !cgroups.PathExists(path) {
+			continue
+		}
+		if err := writeFile(path, CgroupProcesses, strconv.Itoa(pid)); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // Freeze toggles the container's freezer cgroup depending on the state
 // provided
 func (m *Manager) Freeze(state configs.FreezerState) error {
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -350,6 +350,20 @@ func (m *Manager) Set(container *configs.Config) error {
 	panic("not implemented")
 }
 
+func (m *Manager) AddProcess(pid int) error {
+	for name, path := range m.Paths {
+		_, ok := subsystems[name]
+		if !ok || !cgroups.PathExists(path) {
+			continue
+		}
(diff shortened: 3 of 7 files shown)


cgroups/cgroups.go
```diff
@@ -34,6 +34,9 @@ type Manager interface {
 
 	// Set the cgroup as configured.
 	Set(container *configs.Config) error
+
+	// Enters the specified process into these cgroups.
```
vishh: nit: Rephrase as `Enters the specified process into the cgroup set` just to be consistent.

cgroups/systemd/apply_systemd.go
```diff
@@ -350,6 +350,20 @@ func (m *Manager) Set(container *configs.Config) error {
 	panic("not implemented")
 }
 
+func (m *Manager) EnterProcess(pid int) error {
```
vishh: Will it be possible to re-use the code in `fs` here?

vishh:
Overall LGTM. Just a few comments.

rjnagal:
How do we handle cleanup? If the charged container is going away, is it okay to kill the moved processes? Document the behavior somewhere.

cgroups/cgroups.go
vmarmol: Done

cgroups/systemd/apply_systemd.go
vmarmol: Unfortunately it's not easy. We can make a separate function and have it depend on that, but I'm not sure its worth it for a short snippet.

vmarmol:
@rjnagal documented the behavior in the interface. It will be killed alongside the rest of the container.

cgroups/cgroups.go
```diff
@@ -34,6 +34,9 @@ type Manager interface {
 
 	// Set the cgroup as configured.
 	Set(container *configs.Config) error
+
+	// Enters the specified process into the cgroup set.
+	EnterProcess(pid int) error
```
mrunalp: Minor: Should we call this AddProcess? :)

vmarmol: SGTM, made the change :)

cgroups/fs/apply_raw.go
```diff
@@ -161,6 +161,20 @@ func (m *Manager) Set(container *configs.Config) error {
 	return nil
 }
 
+func (m *Manager) AddProcess(pid int) error {
+	for name, path := range m.Paths {
+		_, ok := subsystems[name]
```
LK4D4: isn't `join` function doing same? Also, there is `join` for systemd now too.

vmarmol: `join` also creates the cgroup. Arguably this doesn't need to happen for Add. But it can also be argued that since the directory exists it will work as well. WDYT about making `join` call add?

LK4D4: Yeah, you're right it is different logic. Maybe wrapper with proper name around `writeFile(path...` will help me in future to distinct such cases.

avagin:
@vmarmol  We kill all processes in a cgroup only if a CT is created without its own pidns. Do we need to kill charged processes?
```
        if !c.config.Namespaces.Contains(configs.NEWPID) {
                if err := killCgroupProcesses(c.cgroupManager); err != nil {
                        log.Warn(err)
                }
        }
```

vmarmol:
Closing this for now. The feature I thought it would help, didn't need it in the end. We can revisit if we find a new need in the future. Thanks for the reviews!

PR #504, opened 2015-04-02
Fix pdeathsig and ppid for supervisor running as pid1

crosbymichael:
For some reason when the supervisor of a container is PID1 the parent
death signal was being delivered to the container's init process even if
it did not die.  Also the ppid() will return 1 for a container running
in the pid namespace of it's parent and the parent is pid1.

Fixes: https://github.com/docker/docker/issues/12015


LK4D4:
LGTM

ibuildthecloud:
@crosbymichael @LK4D4 you guys are rock stars, thanks for fixing this.

PR #506, opened 2015-04-03
Change memory usage by minus cache memory

wonderflow:
Fixes [#10824](https://github.com/docker/docker/issues/10824)

Calculate the real usage_memory by "used - buffers cached".

Using cgroup file is  "memory.usage_in_bytes - memory.stat.cache"

Signed-off-by: Sun Jianbo <wonderflow@zju.edu.cn>


crosbymichael:
@wonderflow +1 to what @avagin said,  I think we should add cache to the output then consume that on the client to clarify the results.

wonderflow:
OK I'll add it.

PR #507, opened 2015-04-03
Change nsinit root to /var/run/nsinit

crosbymichael:


mrunalp:
LGTM

SaiedKazemi:
@crosbymichael @avagin 
LGTM.  Although now I understand how it works, it's a good idea to change the default value so others won't be confused.
By the way, I think the Checkpoint() and Restore() APIs will be more flexible if the CRIU images directory was passed in.  If nil is passed in, it can use filepath.Join(c.root, "checkpoint").  But if it's set to a particular path, that would be the images directory to use.

PR #509, opened 2015-04-06
Add value checking on relabel command for selinux

rhatdan:
Currently relabel requires an exact match on passed in relabel string.  This patch will allow the passing of multiple params in the string, and just look for Z or z

### DIFF
--- a/label/label_selinux.go
+++ b/label/label_selinux.go
@@ -104,7 +104,13 @@ func Relabel(path string, fileLabel string, relabel string) error {
 	if fileLabel == "" {
 		return nil
 	}
-	if relabel == "z" {
+	if !strings.ContainsAny(relabel, "zZ") {
+		return nil
+	}
+	if strings.Contains(relabel, "z") && strings.Contains(relabel, "Z") {
+		return fmt.Errorf("Bad SELinux option z and Z can not be used together")
+	}
+	if strings.Contains(relabel, "z") {
 		c := selinux.NewContext(fileLabel)
 		c["level"] = "s0"
 		fileLabel = c.Get()
--- a/label/label_selinux_test.go
+++ b/label/label_selinux_test.go
@@ -87,3 +87,22 @@ func TestDuplicateLabel(t *testing.T) {
 		t.Errorf("DisableSecOpt Failed level incorrect")
 	}
 }
+func TestRelabel(t *testing.T) {
+	testdir := "/tmp/test"
+	label := "system_u:system_r:svirt_sandbox_file_t:s0:c1,c2"
+	if err := Relabel(testdir, "", "z"); err != nil {
+		t.Fatal("Relabel with no label failed: %v", err)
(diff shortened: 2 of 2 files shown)


label/label_selinux.go
```diff
@@ -104,7 +104,13 @@ func Relabel(path string, fileLabel string, relabel string) error {
 	if fileLabel == "" {
 		return nil
 	}
-	if relabel == "z" {
+	if !strings.ContainsAny(relabel, "z & Z") {
+		return nil
+	}
+	if strings.Contains(relabel, "z & Z") {
```
mrunalp: I think this is probably incorrect, unless you are searching for the substring "z & Z" in relabel.

rhatdan: Yes I think you are right, I am looking for ContainersBoth

mrunalp:
You have a merge commit. Could you rebase?

label/label_selinux.go
```diff
@@ -104,7 +104,13 @@ func Relabel(path string, fileLabel string, relabel string) error {
 	if fileLabel == "" {
 		return nil
 	}
-	if relabel == "z" {
+	if !strings.ContainsAny(relabel, "z & Z") {
```
mrunalp: You can do this here otherwise getting a space or & will also satisfy the condition.
```go
if !strings.ContainsAny(relabel, "zZ") {
```

rhatdan:
@mrunalp Must have had my head up my butt when doing the patch originally.  I googled for an example, and this is what I saw.   Fixed.  I also added a small patch to fix the label on /dev/mqueue.  I probably should keep these as two different pull requests.

mrunalp:
@rhatdan Ahh, okay :) Yes, you should create a separate PR for the mqueue fix.

label/label_selinux.go
```diff
@@ -104,7 +104,13 @@ func Relabel(path string, fileLabel string, relabel string) error {
 	if fileLabel == "" {
 		return nil
 	}
-	if relabel == "z" {
+	if !strings.ContainsAny(relabel, "zZ") {
+		return nil
+	}
+	if strings.Contains(relabel, "z") {
```
mrunalp: Might be better to do:
```
                if strings.Contains(relabel, "z") && strings.Contains(relabel, "Z") {
		            return fmt.Errorf("Bad SELinux Option z & Z can not be used together")
                }
                c := selinux.NewContext(fileLabel)
                c["level"] = "s0"
                fileLabel = c.Get()
                return selinux.Chcon(path, fileLabel, true)
```

rhatdan:
Back to the original way I had it, and split out the mqueue fix to a separate patch.

mrunalp:
@rhatdan Thanks! LGTM

mrunalp:
@dqminh Could you take a look?

dqminh:
LGTM

PR #510, opened 2015-04-07
nsinit: Adds support for exec into docker containers.

mrunalp:
This is a POC for nsinit exec support for docker containers.

@crosbymichael @LK4D4 @vmarmol WDYT about adding this? I can clean it up if we are okay adding this.

Signed-off-by: Mrunal Patel

### DIFF
--- a/nsinit/docker.go
+++ b/nsinit/docker.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/codegangsta/cli"
+	"github.com/docker/libcontainer"
+	"github.com/docker/libcontainer/cgroups/systemd"
+	"github.com/docker/libcontainer/configs"
+)
+
+const (
+	dockerRoot    = "/var/run/docker/execdriver/native"
+	stateFilename = "state.json"
+)
+
+func loadDockerFactory(context *cli.Context) (libcontainer.Factory, error) {
+	cgm := libcontainer.Cgroupfs
+	if systemd.UseSystemd() {
+		cgm = libcontainer.SystemdCgroups
+	}
+	return libcontainer.New(dockerRoot, cgm)
+}
+
+func loadState(root string) (*libcontainer.State, error) {
+	f, err := os.Open(filepath.Join(root, stateFilename))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, err
+		}
+		return nil, err
+	}
+	defer f.Close()
+	var state *libcontainer.State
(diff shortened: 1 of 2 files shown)


dqminh:
I like the feature, but perhaps it can be implemented as a flag to normal `exec` since the code looks almost the same except that we load the state file from docker-specific directory. Something like `nsinit exec --container a560... bash`

rjnagal:
Agree w/ @dqminh. Are there going to be differences from exec other than state file directory ?

mrunalp:
@dqminh @rjnagal Yes, it makes sense to add this as a flag to exec. I will make the changes. Thanks!

mrunalp:
Updated.

nsinit/docker.go
```diff
@@ -0,0 +1,52 @@
...
+	return libcontainer.New(dockerRoot, cgm)
+}
+
+func loadState(root string) (*libcontainer.State, error) {
+	f, err := os.Open(filepath.Join(root, stateFilename))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, err
+		}
+		return nil, err
```
dqminh: both returns here is the same. We can probably just have one `return nil, err` instead.

dqminh:
LGTM other than one small nit.

nsinit/docker.go
mrunalp: Fixed.

crosbymichael:
@mrunalp what is the usecase for this?

mrunalp:
A light weight way to exec into docker containers and also a debug tool.

rjnagal:
@crosbymichael WDYT about this one? It does make running exec simpler in some cases.

crosbymichael:
idk, i would rather not add docker specific things too this tool

mrunalp:
I will try to bat one more time :) I think there are requirements cropping up for such a utility. It is similar to how we used to be able to enter into a docker container using its container.json.

mrunalp:
We could add a disclaimer that this is not a supported tool and use at your own risk.

crosbymichael:
Why would you not use docker exec or this?

ncdc:
@crosbymichael we're looking for a way to avoid proxying in/out/err through the docker daemon - trying to minimize hops and potential bottlenecks

dqminh:
I like this too for local exec. One can use `nsinit -docker` to bypass the needs of docker daemon and gives the user more control over the exec sessions ( i.e., wait / stop / kill exec sessions properly without resolving to patch the docker API which we do now ). The user admittedly have to take care of shipping a compatible nsinit version with current docker, but that can be taken care of easily.

LK4D4:
@crosbymichael What your thoughts now?

PR #511, opened 2015-04-07
Add seccomp feature to libcontainer

keloyang:
it makes a simple way to filter system calls, and don't use third-party libraries.

Signed-off-by: Yang Shukui <yangshukui@huawei.com>


LK4D4:
Seems legit.
ping @crosbymichael

PR #512, opened 2015-04-07
Fix share ns from container and the container has shared ns with host

coolljt0725:
Signed-off-by: Lei Jitang <leijitang@huawei.com>

when create a container with `--net=container:parent`, and the net mode of container `parent` 
is `--net=host`, something wrong. Steps to reproduce:
step 1 `docker run --rm -ti --name parent --net=host busybox`
<pre><code>[l00284783@localhost docker]$ docker run --rm -ti --name parent --net=host busybox
/ # ifconfig
docker0   Link encap:Ethernet  HWaddr 00:00:00:00:00:00
          inet addr:172.17.42.1  Bcast:0.0.0.0  Mask:255.255.0.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:1761 errors:0 dropped:0 overruns:0 frame:0
          TX packets:1609 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:156885 (153.2 KiB)  TX bytes:2676950 (2.5 MiB)

ens3f0    Link encap:Ethernet  HWaddr 52:54:00:0E:AF:A0
          inet addr:9.81.1.183  Bcast:9.81.255.255  Mask:255.255.0.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:134326 errors:0 dropped:0 overruns:0 frame:0
          TX packets:71062 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:79910134 (76.2 MiB)  TX bytes:11759548 (11.2 MiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:7112 errors:0 dropped:0 overruns:0 frame:0
          TX packets:7112 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:2819299 (2.6 MiB)  TX bytes:2819299 (2.6 MiB)

/ #</code></pre>
Step2 `docker run --name child --net=container:parent busybox ifconfig`
<pre><code>[l00284783@localhost docker]$ docker run --name child --net=container:parent busybox ifconfig
lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
</code></pre>

The container `child` suppose to have the same net namespace of the container `parent` and apparently its not.

The reason is that when a container is `--net=host`,  `state.NamespacePaths[NEWNET]` is nil
due to the https://github.com/docker/docker/blob/master/daemon/execdriver/native/create.go#L93
has remove `configs.NEWNET` and https://github.com/docker/libcontainer/blob/master/container_linux.go#L306 will not store the net namespace path, so any container want to share net namespace with this container can't work.

I think ipc namespace has same problem, but I not sure, I haven't test it.

I think the ns path should always store in `state.NamespacePaths` no matter it is a private or shared.

I think this patch can solve this.

### DIFF
--- a/container_linux.go
+++ b/container_linux.go
@@ -303,7 +303,12 @@ func (c *linuxContainer) currentState() (*State, error) {
 		CgroupPaths:          c.cgroupManager.GetPaths(),
 		NamespacePaths:       make(map[configs.NamespaceType]string),
 	}
-	for _, ns := range c.config.Namespaces {
+	for _, ns := range []configs.Namespace{
+		{Type: "NEWNS"},
+		{Type: "NEWUTS"},
+		{Type: "NEWIPC"},
+		{Type: "NEWPID"},
+		{Type: "NEWNET"}} {
 		state.NamespacePaths[ns.Type] = ns.GetPath(c.initProcess.pid())
 	}
 	return state, nil


coolljt0725:
I don't know why Jenkins failed ping @cpuguy83 @jfrazelle

dqminh:
@coolljt0725 Jenkins' log is here https://jenkins.dockerproject.com/job/Libcontainer-PRs/348/console

Seems like you need to check for the state's Path when setting the namespaces state. Also can you create an integration test for this ?

coolljt0725:
@dqminh  thanks for your tips. I know its failed here `container_linux_test.go:176: expected path "/networks/fd" but received "/proc/363/ns/net"` 
But I don't understand the test clear.

LK4D4:
@coolljt0725 You now replacing all namespaces in config, you should set pid namespaces only when it isn't listed in `Namespaces` already.

rjnagal:
@coolljt0725 Thanks for the fix. The test is failing because a path is already set in the config and we are ignoring it and reading the actual ns path from system.  When you create the config, you'll need to prefill it with the paths that already exist in the namespaces.

rjnagal:
Does this fix the problem? It seems like docker is passing a net namespace with empty path. And libcontainer will create a namespace whenever the path is empty. In this case, docker should either not set net namespace or set it to the other container that we want to join. Since the other container is in host namespace, we should still land in host's namespace.

vishh:
@LK4D4: Is this fix valid? I think docker should not be passing an empty string as part of Namespaces[configs.NEWNET] whenever a container tries to join another container which is bound to the host namespace [here] (https://github.com/docker/docker/blob/master/daemon/execdriver/native/create.go#L139)

calavera:
I think what @vishh suggests is a better fix than this change. It seems that this fix just ignores every configuration for the container.

LK4D4:
@vishh I think that writing pid-namespace to state is more convenient fix and I'm personally treating this as bug that we don't do this.

LK4D4:
Another argument, that we need add such checks in two places in docker, and only in one place here.

vishh:
@LK4D4: Are we waiting on @coolljt0725 to fix this PR?

LK4D4:
@vishh That's what we do usually. I think we can carry it if it is superimportant.

vishh:
I am assuming we need this for 1.6.

jfrazelle:
yes please :)

PR #513, opened 2015-04-07
Update existing get state test to work with real ns paths.

vishh:
This should help #512 pass.

### DIFF
--- a/container_linux_test.go
+++ b/container_linux_test.go
@@ -120,7 +120,7 @@ func TestGetContainerState(t *testing.T) {
 	var (
 		pid                 = os.Getpid()
 		expectedMemoryPath  = "/sys/fs/cgroup/memory/myid"
-		expectedNetworkPath = "/networks/fd"
+		expectedNetworkPath = fmt.Sprintf("/proc/%d/ns/net", pid)
 	)
 	container := &linuxContainer{
 		id: "myid",


rjnagal:
LGTM.

LK4D4:
I don't see how this is helps. `/fs/network` valid path. And if `Path` is specified for Namespace, then it will be returned from `ns.GetPath`. So this change probably masking possible case when we using arbitrary namespace.

vishh:
@LK4D4: If #512 is a valid fix, this test should be valid as well.

rjnagal:
The test is checking for current state, which should be same as the supplied config in real world, right? The container we are querying is already up and init is pointing to the same config.

LK4D4:
@vishh #512 is invalid fix, it replaces namespaces which was passed with `Path` specified.

vishh:
@LK4D4: Why so? The path that was passed in must match the namespaces of the init process right? Why do we want to support passing arbitrary paths instead of the canonical /proc/pid/ns/ paths?

LK4D4:
@vishh It is good thing to have. For example I can create my namespace with iproute2 and pass it to libcontainer and there is no pid, only path.

vishh:
Ah! I am not aware of that use case. Thanks for clarifying @LK4D4. I will close this PR then?

LK4D4:
@vishh Yup.

PR #514, opened 2015-04-07
Ensure that state always contains pathes to all namespaces

LK4D4:
Thanks coolljit0725 for initial fix.

Closes #512

### DIFF
--- a/configs/namespaces.go
+++ b/configs/namespaces.go
@@ -16,6 +16,17 @@ const (
 	NEWUSER NamespaceType = "NEWUSER"
 )
 
+func NamespaceTypes() []NamespaceType {
+	return []NamespaceType{
+		NEWNET,
+		NEWPID,
+		NEWNS,
+		NEWUTS,
+		NEWIPC,
+		NEWUSER,
+	}
+}
+
 // Namespace defines configuration for each namespace.  It specifies an
 // alternate path that is able to be joined via setns.
 type Namespace struct {
--- a/container_linux.go
+++ b/container_linux.go
@@ -306,5 +306,11 @@ func (c *linuxContainer) currentState() (*State, error) {
 	for _, ns := range c.config.Namespaces {
 		state.NamespacePaths[ns.Type] = ns.GetPath(c.initProcess.pid())
 	}
+	for _, nsType := range configs.NamespaceTypes() {
+		if _, ok := state.NamespacePaths[nsType]; !ok {
+			ns := configs.Namespace{Type: nsType}
+			state.NamespacePaths[ns.Type] = ns.GetPath(c.initProcess.pid())
+		}
+	}
 	return state, nil
 }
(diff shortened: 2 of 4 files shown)


LK4D4:
ping @vishh @rjnagal @crosbymichael

LK4D4:
Sorry, have no idea why it is worked on my host :)

crosbymichael:
lol

crosbymichael:
I don't think that test will work because the destination of the symlink is outside of the container's rootfs and PID namespace for viewing it in proc

integration/exec_test.go
```diff
@@ -574,3 +574,70 @@ func testFreeze(t *testing.T, systemd bool) {
...
+
+	p := &libcontainer.Process{
+		Args: []string{"sleep", "10"},
+		Env:  standardEnvironment,
+	}
+	err = container.Start(p)
+	if err != nil {
+		t.Fatal(err)
+	}
+	p.Signal(os.Kill)
```
crosbymichael: race?

container_linux.go
```diff
@@ -306,5 +306,11 @@ func (c *linuxContainer) currentState() (*State, error) {
 	for _, ns := range c.config.Namespaces {
 		state.NamespacePaths[ns.Type] = ns.GetPath(c.initProcess.pid())
 	}
+	for _, nsType := range configs.NamespaceTypes() {
```
crosbymichael: Would it be better just to do this in one loop?  instead of the one above this and this one below?

integration/exec_test.go
```diff
@@ -574,3 +574,70 @@ func testFreeze(t *testing.T, systemd bool) {
...
+	}
+
+	container, err := factory.Create("test", config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer container.Destroy()
+
+	p := &libcontainer.Process{
+		Args: []string{"sleep", "10"},
```
mrunalp: Maybe use cat so it is deterministic as you can make it exit by closing stdin.

container_linux.go
LK4D4: I though about it. But for me it leads to unnecessary complexity because of array-ness of Namespaces. Maybe you can propose something.

crosbymichael: ok, i think it's fine for now then

LK4D4:
@mrunalp I changed for cat.

crosbymichael:
LGTM

configs/namespaces.go
```diff
@@ -16,6 +16,17 @@ const (
 	NEWUSER NamespaceType = "NEWUSER"
 )
 
+func NamespaceTypes() []NamespaceType {
```
calavera: is this function necessary? it could just be a var, couldn't it?

mrunalp:
LGTM

configs/namespaces.go
crosbymichael: Probably becuase exported vars in go can be mutabed while the return of the a function cannot

crosbymichael: Fun things like doing this show the real issue.

```go
func init() {
    io.EOF = errors.New("hahahahah")
}
```

vishh:
Thanks for the quick fix @LK4D4

PR #516, opened 2015-04-08
Append childpipe for adding addtional Fds to container

crosbymichael:
This takes the work done in #496 and changes this so that we append the childPipe to the end of the processes FD set.  We provide an internal env var for identifying the childPipe's fd to the init process and any of the user defined extra files will be exactly how they specify them.  

Closes #496

### DIFF
--- a/container_linux.go
+++ b/container_linux.go
@@ -139,7 +139,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
-	cmd.ExtraFiles = []*os.File{childPipe}
+	cmd.ExtraFiles = append([]*os.File{childPipe}, p.ExtraFiles...)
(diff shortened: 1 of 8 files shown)


crosbymichael:
@Mic92 i carried your PR and change it a little so that we didn't have to do the complex dup and close logic and it will be better for consumers like yourself as you will not get fd 3 if you add a file instead of libcontainer taking it over and using it.  Let me know what you think and thanks for the PR.  Your commit is included in this one.

LK4D4:
Also not sure how tests passing with that.

crosbymichael:
@LK4D4 `cmd.ExtraFiles` will never be empty becuase we add the child pipe to it before we get the `len()`

LK4D4:
@crosbymichael oh, okay.

mrunalp:
LGTM

LK4D4:
LGTM

Mic92:
Hi, what is about? https://github.com/docker/libcontainer/blob/master/nsenter/nsexec.c#L161
It exptects the pipe at fd=3.

PR #517, opened 2015-04-09
Add support for ARMv8 (AKA: arm64 or Aarch64)

adconrad:
The commit itself should be self-explanatory.  This was tested with docker 1.5 on Ubuntu with gccgo. Should work with golang as well, in theory, but we'll test that theory when the golang port is more complete.

### DIFF
--- a/label/label_selinux.go
+++ b/label/label_selinux.go
@@ -104,7 +104,13 @@ func Relabel(path string, fileLabel string, relabel string) error {
 	if fileLabel == "" {
 		return nil
 	}
-	if relabel == "z" {
+	if !strings.ContainsAny(relabel, "zZ") {
+		return nil
+	}
+	if strings.Contains(relabel, "z") && strings.Contains(relabel, "Z") {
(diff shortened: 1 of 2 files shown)


crosbymichael:
@adconrad code looks good but could you sign your commit?

These commits do not have a proper 'Signed-off-by:' marker:
 - 9d076838d7a540d2dab587ea6b344d3ebf6859bb

Please amend each commit to include a properly formatted DCO marker.

Visit the following URL for information about the Docker DCO:
 https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

adconrad:
On Thu, Apr 09, 2015 at 09:32:56AM -0700, Michael Crosby wrote:
> 
> Visit the following URL for information about the Docker DCO:
> https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

Right, can do.  I probably have some PPC fixes and an ARM fix to
commit as well, so I'll sort all of that later today, make sure
to sign them all, and re-propose the PR.

... Adam

crosbymichael:
@adconrad thanks!

PR #519, opened 2015-04-10
add cgroup subsystem hugetlb

Mashimiao:

### DIFF
--- a/cgroups/fs/hugetlb.go
+++ b/cgroups/fs/hugetlb.go
@@ -1,0 +1,29 @@
+package fs
+
+import (
+	"github.com/docker/libcontainer/cgroups"
+	"github.com/docker/libcontainer/configs"
+)
+
+type HugetlbGroup struct {
+}
+
+func (s *HugetlbGroup) Apply(d *data) error {
+	// we just want to join this group even though we don't set anything
+	if _, err := d.join("hugetlb"); err != nil && !cgroups.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


vmarmol:
LGTM, thanks @Mashimiao!

cgroups/fs/hugetlb.go
```diff
@@ -0,0 +1,29 @@
...
+	return nil
+}
+
+func (s *HugetlbGroup) Remove(d *data) error {
+	return removePath(d.path("hugetlb"))
+}
+
+func (s *HugetlbGroup) GetStats(path string, stats *cgroups.Stats) error {
+	return nil
+}
```
dqminh: should we add hugetlb to the default subsystems to be joined too ?

rjnagal: I don't think its enabled everywhere by default.

crosbymichael:
Shouldn't this be added to non-systemd code also?

Mashimiao:
Thanks @crosbymichael ,  non-systemd added.

cgroups/fs/hugetlb.go
dqminh: oops, i meant that this should be added to the list of subsystems to be applied ( which is fixed now ).

dqminh:
LGTM

crosbymichael:
LGTM

PR #520, opened 2015-04-10
Change mount point propogation to default to slave

rhatdan:
As an admin I would expect that if I volume mount a partition into a container, and then later add a mount point to the volume, it will show up in the container.

### DIFF
--- a/configs/config.go
+++ b/configs/config.go
@@ -37,6 +37,9 @@ type Config struct {
 	// bind mounts are writtable.
 	Readonlyfs bool `json:"readonlyfs"`
 
+	// Privatefs will mount the container's rootfs as private where mount points from the parent will not propogate
+	Privatefs bool `json:"privatefs"`
+
 	// Mounts specify additional source and destination paths that will be mounted inside the container's
 	// rootfs and mount namespace if specified
 	Mounts []*Mount `json:"mounts"`
--- a/rootfs_linux.go
+++ b/rootfs_linux.go
@@ -248,9 +248,9 @@ func mknodDevice(dest string, node *configs.Device) error {
 }
 
 func prepareRoot(config *configs.Config) error {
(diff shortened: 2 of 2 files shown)


rhatdan:
We need this change to allow autofs from the host to work.  Is there a good reason to choose PRIVATE over SLAVE?

ibuildthecloud:
+1

crosbymichael:
@alexlarsson Can you remember the reason why we defaulted to `MS_PRIVATE`?

ibuildthecloud:
@crosbymichael I thought some of this had to do with the fact that by default the host mount namespace will not see the container mounts.  So the host mount namespace doesn't get polluted

ibuildthecloud:
@rhatdan Doesn't something need to change here too https://github.com/rhatdan/libcontainer/blob/slave/rootfs_linux.go#L133

By default the bind mounts are private, are you expecting those bind mounts to be SLAVE also?

rhatdan:
@ibuildthecloud With slave the hosts mnt namespace does not see the containers, but the container can see changes to the host if the hosts volume mounted into the container.

@ibuildthecloud The bind mounts are the ones passed in as being private, I believe.  

I tested this out with a docker run -v /usr:/usr fedora /bin/sh,
Then mounted a directory on /usr/share/test, and the container could see the contents.

ibuildthecloud:
@rhatdan thanks, let me test this

crosbymichael:
LGTM

ibuildthecloud:
All my tests are good.  This patch has minimal impact if your host mounts are set to MS_PRIVATE.

mrunalp:
LGTM

jeremyeder:
Cool, only 10 days late :-) @rhatdan asked us to test this; we found no change in container create/destroy performance between docker-1.5 and docker-1.6.

mrunalp:
@jeremyeder Good to know :)

PR #523, opened 2015-04-10
Read _LIBCONTAINER_INITPIPE in nsexec.c

LK4D4:
This fixes ExtraFiles for execin
Thanks @Mic95 for report

### DIFF
--- a/integration/execin_test.go
+++ b/integration/execin_test.go
@@ -314,3 +314,78 @@ func TestExecInEnvironment(t *testing.T) {
 		t.Fatalf("unexpected running process, output %q", out)
 	}
 }
+
+func TestExecinPassExtraFiles(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+	rootfs, err := newRootfs()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


crosbymichael:
LGTM

integration/execin_test.go
```diff
@@ -314,3 +314,78 @@ func TestExecInEnvironment(t *testing.T) {
...
+		ExtraFiles: []*os.File{pipein1, pipein2},
+		Stdin:      nil,
+		Stdout:     &stdout,
+	}
+	err = container.Start(inprocess)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	waitProcess(inprocess, t)
```
dqminh: should we wait for the original process (i.e., `cat`) here too ?

mrunalp:
LGTM

integration/execin_test.go
LK4D4: I think it is okay, it will be die on defer and not crucial for test.

dqminh:
LGTM

PR #528, opened 2015-04-13
WIP: Mounting cgroups readonly into containers

mrunalp:
This is a PoC / RFC to explore if this is worth doing. (I have only tested this with systemd and will add fs support/refactor if we think this is useful).

We only mount the container's own cgroups under /cgroups.

### DIFF
--- a/init_linux.go
+++ b/init_linux.go
@@ -40,15 +40,16 @@ type network struct {
 
 // initConfig is used for transferring parameters from Exec() to Init()
 type initConfig struct {
-	Args             []string        `json:"args"`
-	Env              []string        `json:"env"`
-	Cwd              string          `json:"cwd"`
-	Capabilities     []string        `json:"capabilities"`
-	User             string          `json:"user"`
-	Config           *configs.Config `json:"config"`
-	Console          string          `json:"console"`
-	Networks         []*network      `json:"network"`
-	PassedFilesCount int             `json:"passed_files_count"`
+	Args             []string         `json:"args"`
+	Env              []string         `json:"env"`
+	Cwd              string           `json:"cwd"`
+	Capabilities     []string         `json:"capabilities"`
+	User             string           `json:"user"`
+	Config           *configs.Config  `json:"config"`
+	Console          string           `json:"console"`
+	Networks         []*network       `json:"network"`
+	PassedFilesCount int              `json:"passed_files_count"`
+	CgroupMounts     []*configs.Mount `json:"cgroup_mounts"`
 }
 
 type initer interface {
--- a/nsinit/exec.go
+++ b/nsinit/exec.go
@@ -23,6 +23,7 @@ var execCommand = cli.Command{
 	Action: execAction,
 	Flags: append([]cli.Flag{
 		cli.BoolFlag{Name: "tty,t", Usage: "allocate a TTY to the container"},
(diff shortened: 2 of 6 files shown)


mrunalp:
@rhatdan ping

avagin:
What is a problem you solve here? If you are going to run systemd in CT, it isn't enough
```
Either pre-mount all cgroup hierarchies in full into the container, or leave that to systemd which will do so if they are missing. Note that it is explicitly not OK to just mount a sub-hierarchy into the container as that is incompatible with /proc/$PID/cgroup (which lists full paths). Also the root-level cgroup directories tend to be quite different from inner directories, and that distinction matters. It is OK however, to mount the "upper" parts read-only of the hierarchies, and only allow write-access to the cgroup subtree the container runs in.
```

mrunalp:
This will allow container processes to query their cgroup limits. For e.g. JVM tuning could be done looking at memory limits.

rhatdan:
I ased @mrunalp to look into getting this to work.  There are really two goals here.  

* Setup the Minimal amount of information systemd needs to be able to run within the container.  We want to see how easily we could fool systemd to run with a /sys/fs/cgroup hierarchy mounted in Read/Only.  as @avagin states we might need to mount in other parts of the the cgroup hierarchy to get this to work.  But with this patch we can start experimenting.

Secondly there are big discussions going on in Red Hat bugzillas as well as on github and other places on getting tools running inside a container (cgroup) to be able to answer fundamental questions like "how much memory can I attempt take advantage of?"  

Currently tools like free/top etc are only looking at content in /proc which is giving them total memory on the system, and the tools like jvm's are being fooled.  If we leak the cgroup information into the container then the tools could become more intelligent.  

But mounting the entire /sys/fs/cgroup into the container, could be considered an information leak.  This patch would allow us to only mount the cgroup for the container into the container.

avagin:
@rhatdan @mrunalp Thank you for the explanation.

hqhq:
@rhatdan I'm interested in that too, getting tools like these running and working inside a container is very important and useful. 

This PR is a work around, I think libvirt already did something like this. But we still need to change tools so they will use info from cgroup instead of /proc.

As discussed in https://github.com/docker/docker/issues/8427, I hope we can have a more functional solution, so we don't need to change our tools.

crosbymichael:
What do you all think about this approach?

https://github.com/docker/libcontainer/compare/master...crosbymichael:cgroup-mount

the problem is where in the container we mount cgroups because /sys is RO and we cannot create the root dirs to even start to bind mount

rhatdan:
If we mount a tmpfs/rw at /sys/fs/cgroup or /sys/fs/cgroup from the host on /sys/fs/cgroup then the tools inside the container look for the cgroup file system in it's standard path.  I am not crazy about creating a new non standard path at /cgroup.

crosbymichael:
Ya, tmpfs mount on `/sys/fs/cgroup` sounds like the best plan

crosbymichael:
Ok, i updated my branch here to support /sys/fs/cgroup with a tmpfs mount

https://github.com/docker/libcontainer/compare/master...crosbymichael:cgroup-mount

rhatdan:
Why hard code /sys/fs/cgroup.  There is a function call to get that path.

mrunalp:
@crosbymichael Nice!

PR #531, opened 2015-04-14
integration: wait all test processes

avagin:
Otherwise a container cannot be destroyed and we will get an error
in the next case:
go test -v .
=== RUN TestExecPS
--- FAIL: TestExecPS (0.02s)
        exec_test.go:43: <nil>: [0] Id already in use: Container with id exists: testCT

Cc: @LK4D4 


mrunalp:
LGTM

crosbymichael:
LGTM

PR #532, opened 2015-04-14
check "/sbin/apparmor_parser" in apparmor.IsEnabled()

tifayuki:
In ubuntu 14.04, Apparmor is installed by default. When uninstalling apparmor with `apt-get purge apparmor`, `/sbin/apparmor_parser` will be delete. However, `/sys/kernel/security/apparmor` and `/sys/module/apparmor/parameters/enabled` remains on the host, resulting in the following error
```
FATA[0000] Error loading docker apparmor profile: fork/exec /sbin/apparmor_parser: no such file or directory () 
```

### DIFF
--- a/apparmor/apparmor.go
+++ b/apparmor/apparmor.go
@@ -14,8 +14,10 @@ import (
 
 func IsEnabled() bool {
 	if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil && os.Getenv("container") == "" {
-		buf, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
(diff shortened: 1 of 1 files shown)


crosbymichael:
Just for my own personal info, why do you have apparmor enabled but uninstalled the tools so that profiles cannot be loaded?

tifayuki:
It is one of our user's nodes. No idea why he uninstalled apparmor tools :-/

crosbymichael:
@tifayuki ok, thanks.

LGTM

mrunalp:
LGTM

PR #534, opened 2015-04-16
cr: prepare a container root

avagin:
CRIU has following requirements for a container root:
* it is a mount point
* a parent mount point isn't over-mounted


huikang:
Hi, Andraw,
The patch solved the problem. Thanks.

- Hui

crosbymichael:
LGTM

PR #535, opened 2015-04-16
Adds support for setting system properties.

mrunalp:
This allows one to do the equivalent of sysctl -w ip.ipv4.ip_forward 1 or any such property.

I will add tests and other bits if we are okay adding this.

### DIFF
--- a/rootfs_linux.go
+++ b/rootfs_linux.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 	"syscall"
@@ -386,4 +387,11 @@ func maskFile(path string) error {
 		return err
 	}
 	return nil
+}
+
+// writeSystemProperty writes the value to a path under /proc/sys as determined from the key.
+// For e.g. net.ipv4.ip_forward translated to /proc/sys/net/ipv4/ip_forward.
+func writeSystemProperty(key, value string) error {
+	keyPath := strings.Replace(key, ".", "/", -1)
+	return ioutil.WriteFile(path.Join("/proc/sys", keyPath), []byte(value), 0644)
 }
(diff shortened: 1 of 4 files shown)


hqhq:
Seems risky if we allow users to set all system properties.

mrunalp:
@hqhq docker could filter what is allowed. In libcontainer, it makes sense to be flexible.

hqhq:
Right, +1 for me then, thanks.

LK4D4:
Looks cool

mrunalp:
Updated with tests.

mrunalp:
@LK4D4 @crosbymichael PTAL.

crosbymichael:
LGTM

rootfs_linux.go
```diff
@@ -387,3 +388,12 @@ func maskFile(path string) error {
 	}
 	return nil
 }
+
+// writeSystemProperty writes the value to a path under /proc/sys as determined from the key.
+// For e.g. net.ipv4.ip_forward translated to /proc/sys/net/ipv4/ip_forward.
+func writeSystemProperty(key, value string) error {
+	path_prefix := "/proc/sys"
```
crosbymichael: camelCase

rootfs_linux.go
```diff
@@ -387,3 +388,12 @@ func maskFile(path string) error {
...
+	key_path := strings.Replace(key, ".", "/", -1)
+	full_path := path.Join(path_prefix, key_path)
```
crosbymichael: Could probably reduce the amount of vars in this func as well

rootfs_linux.go
mrunalp: Fixed.

rootfs_linux.go
mrunalp: Fixed.

LK4D4:
LGTM

crosbymichael:
LGTM

PR #536, opened 2015-04-17
cr: don't remove images after restore

avagin:
This code was added when images were saved in a factory root directory and in
addition this directory was used as a flag.

Now an user can decide when images can be removed.

Cc: @boucher


mrunalp:
LGTM

boucher:
Thanks for the heads up

crosbymichael:
LGTM

PR #538, opened 2015-04-17
fix freeze systemd test

hqhq:
Made a mistake before, freeze test doesn't use newContainer,
systemd test doesn't actually work.


crosbymichael:
LGTM

PR #541, opened 2015-04-17
Mount /sys/fs/cgroup into the container

rhatdan:
Then mount all of the config directories specific to the container.

This pull request grabs some of the ideas from https://github.com/docker/libcontainer/pull/528 and mounts them at the standard location under /sys/fs/cgroup.  

One of my goals is to get systemd to run within a container without having to to a -v /sys/fs/cgroup:/sys/fs/cgroup:ro.

Up til now I have not figured how to fool it without mounting the file system in.  The only problem I see with mounting it in is a leak of information about other processes on the system.

### DIFF
--- a/init_linux.go
+++ b/init_linux.go
@@ -40,15 +40,16 @@ type network struct {
 
 // initConfig is used for transferring parameters from Exec() to Init()
 type initConfig struct {
-	Args             []string        `json:"args"`
-	Env              []string        `json:"env"`
-	Cwd              string          `json:"cwd"`
-	Capabilities     []string        `json:"capabilities"`
-	User             string          `json:"user"`
-	Config           *configs.Config `json:"config"`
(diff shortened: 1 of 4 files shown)


avagin:
A few useful links about systemd in CT:
https://lwn.net/Articles/624767/
http://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
http://lists.freedesktop.org/archives/systemd-devel/2015-February/028772.html

rhatdan:
Yes I am very familiar with these articles/problems, and have talked to Lennart fairly regularly.

dqminh:
I think this should be closed by https://github.com/docker/libcontainer/pull/553

rhatdan:
Yes although this does not satisfy the need for /sys/fs/cgroup mounted into systemd based containers.
I will have to work on separate patches for that.

PR #542, opened 2015-04-18
Set the seed when randMacAddr

sayuan:
If the seed was not seted, all machines get the same MAC address.
Then technique like Open vSwitch won't work due to the same MAC address
of all machines.

### DIFF
--- a/netlink/netlink_linux.go
+++ b/netlink/netlink_linux.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"sync/atomic"
 	"syscall"
+	"time"
 	"unsafe"
 )
 
@@ -53,6 +54,8 @@ type ifreqFlags struct {
 }
 
 var native binary.ByteOrder
+
+var rnd = rand.New(rand.NewSource(time.Now().UnixNano()))
 
 func init() {
 	var x uint32 = 0x01020304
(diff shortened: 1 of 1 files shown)


netlink/netlink_linux.go
```diff
@@ -1213,6 +1214,7 @@ func AddToBridge(iface, master *net.Interface) error {
 }
 
 func randMacAddr() string {
+	rand.Seed(time.Now().UTC().UnixNano())
```
thaJeztah: Should it be seeded for each call, or once? (A similar discussion was in docker recently; https://github.com/docker/docker/pull/12214)

sayuan:
Thank you for the advice. Now I've updated it.

crosbymichael:
@estesp Is this ok or do you think we should do a package level var from `rand.New()` instead of using `rand.Seed`?

estesp:
@crosbymichael I think a package level var is preferred as this is setting the seed for all users of `rand` in the process space.  It is possible some other package/`init()` has set the seed already or, alternatively, is using a different seed and gets init'ed after this package.

@sayuan an example of `rand.New()` is here: https://github.com/docker/docker/blob/master/pkg/namesgenerator/names-generator.go#L312

sayuan:
Please review it again. :)

LK4D4:
LGTM

PR #544, opened 2015-04-19
Exclude the root path from mount point paths

boucher:
This change was necessary for me to get criu/libcontainer running inside of Docker.


boucher:
Sorry, mistakenly included the other branches commit as well.

PR #548, opened 2015-04-21
WIP: Support chroot with no NEWNS

ibuildthecloud:
**NOTE: This code is basically just a proof of concept.  If the basic approach seems fine I'll produce better code.**

This PR is the libcontainer changes needed to support `--mnt=host` in Docker.  That's right, using the the host mnt namespace.  It might seems crazy, but basically what I'm proposing is that in that situation you do a chroot and not a new mnt namespace.  Before I go into the details of this PR let me first explain why I want this.  That way if this approach is way off then maybe a better one can be proposed.

## Motivation

What I want to accomplish is to allow a container to manage mounts in the host mnt namespace.  The primary use case is to support something like ZFS, LVM, or Ceph that requires userspace utilities to manage.  In RancherOS we have no host system and thus managing such subsystems must be done from a container.  With the recent changes in libcontainer to default to SLAVE we can propagate mount from one container to another through the host using this PR.

My first attempt at this functionality was to just use shared mounts.  I ran into the fundamental issue in that when the container mount namespace is removed all mounts are umounted causing the shared mounts to be removed from the host.  This is no good because we can never restart the container managing the mounts.

## How this PR works

NoPivotRoot is removed and replace with a flag for Chroot.  Here's how Chroot and NEWNS work with each other

NEWNS | Chroot | Behavior | Docker possible?
------|--------|-------|------
NEWNS | false | The default Docker behavior when creating a container | Yep
 - | false | No mount namespace and setupRootFs is not called, so no root | Nope
NEWNS | true | New mnt namespace but we don't pivot_root but instead MS_MOVE and chroot | Yes with DOCKER_RAMDISK=true
- | true | No new mnt namespace but setupRootFs is still called and all mounts are setup | Yep,`--mnt=host`

### Cleanup

When a container is deleted and there is no NEWNS and chroot is true, we must manually remove mounts.  In this situation we traverse mounts in the container and manually umount them all.  **One difference is that we first make the propagation private before we umount.**  This way the mounts are not removed from the host.

fyi @lukemarsden

### DIFF
--- a/configs/config.go
+++ b/configs/config.go
@@ -19,10 +19,6 @@ type IDMap struct {
 type Config struct {
 	// NoPivotRoot will use MS_MOVE and a chroot to jail the process into the container's rootfs
 	// This is a common option when the container is running in ramdisk
-	NoPivotRoot bool `json:"no_pivot_root"`
-
-	// ParentDeathSignal specifies the signal that is sent to the container's process in the case
-	// that the parent process dies.
 	ParentDeathSignal int `json:"parent_death_signal"`
 
 	// PivotDir allows a custom directory inside the container's root filesystem to be used as pivot, when NoPivotRoot is not set.
@@ -96,6 +92,11 @@ type Config struct {
 	// ReadonlyPaths specifies paths within the container's rootfs to remount as read-only
 	// so that these files prevent any writes.
 	ReadonlyPaths []string `json:"readonly_paths"`
+
+	// Chroot indicates that the container's process should be chrooted to the Rootfs
+	// If a mount namespace is created this means that we prefer to chroot rather than
+	// pivot_root.  If there is no mount namespace a regular Chroot will be done.
+	Chroot bool `json:"chroot"`
 }
 
 // Gets the root uid for the process on host which could be non-zero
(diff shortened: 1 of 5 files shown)


crosbymichael:
I'm not excited about manually cleaning up mounts

ibuildthecloud:
I'm not either, but it follows the same paradigm as the PID NS.  When PID host was done it also required manual cleanup.

ibuildthecloud:
@avagin yes, as I mentioned before, this was just a POC.  I'm not sure if the approach would even fly with the maintainers, so I'm mostly concerned that we agree on the approach.  I'll fix the code and write proper tests if I get a design ok thumbs up.

ibuildthecloud:
@crosbymichael creating a container with no mnt namespace might be too much madness.  I'm going to pursue the route of see what I can accomplish with persistent mnt namespaces.  I think at the end of the day i'm going to hit a wall where I will need to be able to do `-v /host:/foo` and not have it be automatically slave.  (so no MS_REC on the root).  Not sure how I'll accomplish that.

ibuildthecloud:
I don't think libcontainer is quite ready for this level of crazy (yet).  Closing PR

PR #543, opened 2015-04-19
Add support for providing options to CRIU.

boucher:
In order to do more complex things with checkpointing and restoring of containers it's necessary to have control over where the image files are being saved and whether or not to kill the running process. It's possible more flags will be wanted in the future. Some things probably should always be auto-configured by libcontainer though.

I'd be glad to hear any suggestions for alternate solutions here. This is my first contribution.

PR #551, opened 2015-04-22
Allow dumped image to be transferred to a criu page server

huikang:
--page-server to specify the IP address of criu page server
  --port to specify the port of the criu page server

Signed-off-by: Hui Kang <hkang.sunysb@gmail.com>

### DIFF
--- a/container.go
+++ b/container.go
@@ -115,7 +115,7 @@ type Container interface {
 	//
 	// errors:
 	// Systemerror - System error.
-	Checkpoint(string) error
+	Checkpoint(string, string, string) error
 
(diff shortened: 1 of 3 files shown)


boucher:
Would be nice to combine this with #543

crosbymichael:
Yes, I would vote for having a struct as input instead of many parameters

huikang:
Me too.
I like the idea of using criuOpts *CriuOpts as the parameter for checkpoint
and restore.

- Hui

huikang:
Hi, Michael,
Maybe you could merge #543 and then I will re-submit a patch for
page-server and port.

- Hui

PR #552, opened 2015-04-23
Allow dumped image to be transferred to a criu page server

huikang:
--page-server to specify the IP address of criu page server
--port to specify the port of the criu page server

Docker-DCO-1.1-Signed-off-by: Hui Kang <hkang.sunysb@gmail.com>

### DIFF
--- a/criu_opts.go
+++ b/criu_opts.go
@@ -1,10 +1,16 @@
 package libcontainer
 
+type CriuPageServerInfo struct {
+	Address string // IP address of CRIU page server
+	Port    int32  // port number of CRIU page server
+}
+
 type CriuOpts struct {
-	ImagesDirectory         string // directory for storing image files
-	WorkDirectory           string // directory to cd and write logs/pidfiles/stats to
-	LeaveRunning            bool   // leave container in running state after checkpoint
-	TcpEstablished          bool   // checkpoint/restore established TCP connections
-	ExternalUnixConnections bool   // allow external unix connections
-	ShellJob                bool   // allow to dump and restore shell jobs
+	ImagesDirectory         string              // directory for storing image files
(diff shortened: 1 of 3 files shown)


huikang:
This is a resubmission of  #551 as suggested by boucher.

criu_opts.go
```diff
@@ -1,10 +1,16 @@
...
-	TcpEstablished          bool   // checkpoint/restore established TCP connections
-	ExternalUnixConnections bool   // allow external unix connections
-	ShellJob                bool   // allow to dump and restore shell jobs
+	ImagesDirectory         string              // directory for storing image files
+	WorkDirectory           string              // directory to cd and write logs/pidfiles/stats to
+	LeaveRunning            bool                // leave container in running state after checkpoint
+	TcpEstablished          bool                // checkpoint/restore established TCP connections
+	ExternalUnixConnections bool                // allow external unix connections
+	ShellJob                bool                // allow to dump and restore shell jobs
+	Ps                      *CriuPageServerInfo // allow to dump to criu page server
```
crosbymichael: Maybe `net.Addr` is a better data structure for this? idk what do you think?

huikang:
CriuPageServerInfo is a structure from criurpc/criurpc.pb.go and the same as in criu rpc.
Ross defined the CriuOpts struct. @boucher, what do you think?

crosbymichael:
Ok, that is fine then

boucher:
I have no real opinion. It's just a placeholder for shuttling data around though, so perhaps it makes sense not to have to convert to and from another type.

huikang:
I agree. Thanks for your comments.

- Hui

huikang:
@avagin ok, I will change to that cli format. Thanks.

crosbymichael:
LGTM

PR #554, opened 2015-04-24
Split namespace syscall content for building on non-Linux

estesp:
libcontainer/configs is used by the docker user namespace proposed
patchset to use IDMap for uid/gid maps across the codebase.  Given the
client uses some of this code, it needs to build on non-Linux.  This
separates out the Linux-only syscalls using build tags.

Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com> (github: estesp)

### DIFF
--- a/configs/namespaces.go
+++ b/configs/namespaces.go
@@ -1,9 +1,6 @@
 package configs
 
-import (
-	"fmt"
-	"syscall"
-)
+import "fmt"
 
 type NamespaceType string
 
@@ -32,10 +29,6 @@ func NamespaceTypes() []NamespaceType {
 type Namespace struct {
 	Type NamespaceType `json:"type"`
 	Path string        `json:"path"`
-}
-
-func (n *Namespace) Syscall() int {
-	return namespaceInfo[n.Type]
 }
 
 func (n *Namespace) GetPath(pid int) string {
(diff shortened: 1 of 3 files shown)


dqminh:
LGTM

crosbymichael:
What does the client use?

estesp:
pkg/archive

crosbymichael:
what does that depend on?

estesp:
needs def'n for `config.IDMap` to handle optional remapping on tar/untar/unpack, etc.  Trying to split out client usage of TarWithOptions and Untar from pkg/archive looks very tricky, even though client (today) doesn't need to use the optional IDMap(s) for what it is doing.

crosbymichael:
Humm, i don't think an archive package needs to depend on a libcontainer package.  Seems like it's not the right depedency

estesp:
I need []configs.IDMap all over docker, and of course, as you know, the archiver has to have those maps to do remapping on layer untar.  If that is something we can compartmentalize another way, but given configs and user are not dependent on the rest of libcontainer they seem like reasonable "pieces" to rely on for shared definitions.  If there is a better way, I'm game to do something else.

dqminh:
Even if they dont have to share `libcontainer/configs` ( probably just create their own structure and remap to libcontainer at the execdriver level is better ), I still think that spliting syscall support to `_linux` in config is an ok thing to do.

However this will probably make libct branch not working because of the build tags.

estesp:
@crosbymichael let me know if you want me to have a Docker "shadow" `IDMap` definition that gets cast to libcontainer `config.IDMap` when needed in the execdriver (as @dqminh mentioned above).  That touches a lot of places and I would be happy to refactor the whole user namespace PR, but would like your opinion so I can go one way or the other.

crosbymichael:
I would rather have this struct not in the code of docker right.  So if you could make the type defined there that would be great.

crosbymichael:
Anyways, this is a good idea to do.

LGTM

PR #555, opened 2015-04-27
cgroups/systemd: remove useless code

avagin:
I think the remove code and devices.Set do the same things.


crosbymichael:
LGTM

PR #556, opened 2015-04-27
Remove double cgroup manager initialization.

calavera:
The groups manager should always be provided by the manager.
It looks like this already happened except in the `nsinit init` case.

/cc @crosbymichael

Signed-off-by: David Calavera <david.calavera@gmail.com>


crosbymichael:
Idk, if it was required it should probably be a required parameter.  I don't think it hurts anything to create this type does it?

mrunalp:
Agree with @crosbymichael

calavera:
It still feels weird, but I agree that it doesn't hurt :neutral_face:

PR #557, opened 2015-04-27
Don't read or store the stdio file descriptors in the container config.

boucher:
Instead, write the fd info to the image directory during a checkpoint, and restore with the written info from a given image directory. Also, write completely root independent ext_mount info. These changes enable checkpointing in one container and then restoring into a completely new container.

Docker-DCO-1.1-Signed-off-by: Ross Boucher <rboucher@gmail.com> (github: boucher)


boucher:
ping @avagin if you have a minute to review

avagin:
Originally we save fds before starting an init process, because the init process can change these descriptors. I think we need to preserve this logic. So I think we need to save std_fds into the container state file and then save these std_fds into the image directory.

boucher:
Do you mean, save the fds before the container starts entirely? I can make that change. I don't really know how these things can change over time, but why is it not the case that you'd want to track the state at the same time the checkpoint is made?

avagin:
Because we want to track file descriptors which were provided by an executor. We can restore only them.

avagin:
So we don't need to remove saveStdPipes(), it's called in right places.

boucher:
Would you then recommend continuing to save them in the Config and then writing out the file at checkpoint time? Otherwise I guess we could write a file and make a copy of it at checkpoint time.

avagin:
I think it's the good idea to write them into a file.

boucher:
How about this version?

crosbymichael:
Should't these go in State and not Config?

boucher:
Probably they should, though I think that means adding State to initConfig so we can access it in process_linux start()

container_linux.go
```diff
@@ -528,13 +556,6 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 		log.Warn(st.String())
 	}()
 
-	if process != nil {
```
boucher: @avagin It's not clear to me if this is actually needed, but if it is I will put it back.

avagin:
@boucher Could you cherry-pick this commit https://github.com/avagin/libcontainer/commit/19cf489a317bba4f0a26fe0ac11565815661156f

I can't to createa pull-request into your repo.

avagin:
And there are two more commits. You can cherry-pick them too or I will create a separate pull request.
https://github.com/avagin/libcontainer/commit/a4a8783735e9cf15a6dc416e5804520d1ffbd916
https://github.com/avagin/libcontainer/commit/3d5fc968e8603b31a7db074dd6fc692a1efddc18

boucher:
done

avagin:
LGTM

boucher:
@crosbymichael look ok?

container_linux.go
```diff
@@ -469,14 +487,23 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
...
 		}
 	}
-	// Pipes that were previously set up for std{in,out,err}
-	// were removed after checkpoint.  Use the new ones.
-	var i int32
-	for i = 0; i < 3; i++ {
-		if s := c.config.StdFds[i]; strings.Contains(s, "pipe:") {
+
+	var fds []string
+	fdJSON, err := ioutil.ReadFile(filepath.Join(criuOpts.ImagesDirectory, "std_fds.json"))
```
crosbymichael: maybe name this `descriptors.json`

container_linux.go
```diff
@@ -469,14 +487,23 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
...
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(fdJSON, &fds)
```
crosbymichael: We usually do these error checks on one line

container_linux.go
```diff
@@ -361,11 +362,23 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
...
 	}
 
+	// Write the FD info to a file in the image directory
+
+	fdsJSON, err := json.Marshal(c.initProcess.externalDescriptors())
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(filepath.Join(criuOpts.ImagesDirectory, "std_fds.json"), fdsJSON, 0655)
```
crosbymichael: one line for the error check here too

container_linux.go
crosbymichael: Probably move this filename to a `const` in the package

crosbymichael:
@boucher CI failed?

boucher:
Looking into it

crosbymichael:
LGTM

PR #559, opened 2015-04-30
Additional groups lookup

mrunalp:
This PR modifies AdditionalGroups to be a string array and the lookup to translate to group ids is performed in the container as discussed in https://github.com/docker/docker/pull/10717

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -348,3 +348,52 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 
 	return user, nil
 }
+
+// GetAdditionalGroupsPath is a wrapper for GetAdditionalGroups. It reads data from the
+// given file path and uses that data as the arguments to GetAdditionalGroups.
+func GetAdditionalGroupsPath(additionalGroups []string, groupPath string) ([]int, error) {
+	var groupIds []int
+
+	for _, ag := range additionalGroups {
+		groupReader, err := os.Open(groupPath)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to open group file: %v", err)
+		}
+		defer groupReader.Close()
+
+		groupId, err := GetAdditionalGroup(ag, groupReader)
+		if err != nil {
+			return nil, err
+		}
+		groupIds = append(groupIds, groupId)
+	}
+
+	return groupIds, nil
+}
+
+// GetAdditionalGroup looks up the specified group in the passed groupReader.
+func GetAdditionalGroup(additionalGroup string, groupReader io.Reader) (int, error) {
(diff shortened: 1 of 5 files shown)


mrunalp:
@tianon PTAL :)

user/user.go
```diff
@@ -348,3 +348,52 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+// given file path and uses that data as the arguments to GetAdditionalGroups.
+func GetAdditionalGroupsPath(additionalGroups []string, groupPath string) ([]int, error) {
+	var groupIds []int
+
+	for _, ag := range additionalGroups {
+		groupReader, err := os.Open(groupPath)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to open group file: %v", err)
+		}
+		defer groupReader.Close()
```
dqminh: i think `groupPath` can be opened once and reused after.

user/user.go
```diff
@@ -348,3 +348,52 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+
+		groupId, err := GetAdditionalGroup(ag, groupReader)
```
dqminh: Should `GetAdditionalGroup` accepts a list of groups too so it doesnt have to re-parse the group file for each group ? Or maybe we can merge `GetAdditionalGroup` into this function so it's more obvious that we will only read the group file once.

user/user.go
```diff
@@ -348,3 +348,52 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+		if err != nil {
+			return nil, err
+		}
+		groupIds = append(groupIds, groupId)
+	}
+
+	return groupIds, nil
+}
```
dqminh: I think we should also have an unit test for this.

mrunalp: Yes, will add one.

user/user.go
mrunalp: I went down that path but realized that the existing filter won't work for it. I can however create a new filter that checks for multiple groups at the same time so we can avoid multiple passes through the group reader.

dqminh: I think adding a filter that works with multiple groups sounds good.

LK4D4:
ping @mrunalp :)
Would be nice to have docker/docker#10717 in next release

mrunalp:
Am out sick :( hopefully next week

dqminh:
@mrunalp I have some free time ! I can send you some improvements of this if you dont mind.

mrunalp:
Sure, go for it. Thanks.

LK4D4:
Closing in favor of #603

PR #560, opened 2015-05-01
integration: don't create a factory for each test case

avagin:


vendor/src/github.com/docker/libcontainer
```diff
@@ -0,0 +1 @@
+../../../../
```
dqminh: hmm i think we dont need this. If the desired state is to have a compiled nsinit binary, there's `make binary` which installed nsinit binary into `./bundle`

avagin: @dqminh I don't want to execute a docker container to compile nsinit or execute tests. Why is it bad to have this link?

avagin: @dqminh I agree that this commit should not be in this series. I've removed it.

LK4D4:
LGTM

mrunalp:
LGTM

PR #565, opened 2015-05-02
criu path should be a global variable, if it exists.

huikang:
Remove criu path in container struct

Docker-DCO-1.1-Signed-off-by: Hui Kang <hkang.sunysb@gmail.com>

### DIFF
--- a/container_linux.go
+++ b/container_linux.go
@@ -27,7 +27,6 @@ type linuxContainer struct {
 	initPath      string
 	initArgs      []string
 	initProcess   parentProcess
-	criuPath      string
 	m             sync.Mutex
 }
 
@@ -268,7 +267,7 @@ func addArgsFromEnv(evar string, args *[]string) {
 func (c *linuxContainer) checkCriuVersion() error {
 	var x, y, z int
 
-	out, err := exec.Command(c.criuPath, "-V").Output()
+	out, err := exec.Command(CriuPath, "-V").Output()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


huikang:
@avagin , if you have a minute to review this change. Thanks.

crosbymichael:
Why do you think this should be a global?

huikang:
@crosbymichael Criu is typically installed system-wide, instead of for per container. So it looks better to separate the criu path from the container struct. Am I right?

avagin:
My opinion is that it should not be a global variable, because we already have $PATH for that.

huikang:
@avagin Thanks for your comment.
Actually the purpose of this commit is to serve as the first step to de-couple the two criu-related functions, i.e., criuNotifications( ) and criuSwrk ( ) from the container_linux.go. Such that more criu features can be enabled with nsinit, e.g., some criu command (such as page-server) does not require a container factory. I have a separate PR for that if you think this is worthwhile.

avagin:
@huikang For me it isn't clear why we may want to execute criu page-server from nsinit. Could you describe use-cases?

huikang:
Hi, @avagin 
I was initially attempt to implement the container migration for the nsinit command; then I found that some criu-specific function like criuswrk is tied to the container interface.

Regarding to the migration, my thought is as follows.
On the destination host, I would like to run nsinit as a criu page-server (e.g., nsinit ... restore-remote) to receive the dump image and then create a container factory. The nsinit on the destination host waits until some remote source host run "nsinit ... checkpoint --dst IP of the dst". Then the container will be restored on the destination host.
What do you think? Thanks.

crosbymichael:
It's done the way it is now not because we want a different version of criu per container but to remove the use of package level global vars.

huikang:
@crosbymichael 
This makes sense. Then maybe a criu object is better than tie the criu path to individual container object.

avagin:
@huikang Have you seen the p.haul project https://github.com/xemul/p.haul. If you have not, I recommend you to do this. I think it can help you to solve your task. You need to implement a driver for nsinit, other logic about online-migration is already there.

huikang:
@avagin thanks for bringing this to my attention. I will definitely look at this.

PR #567, opened 2015-05-04
hugetlb: Add support of Set and GetStats function

Mashimiao:

### DIFF
--- a/cgroups/fs/hugetlb.go
+++ b/cgroups/fs/hugetlb.go
@@ -3,6 +3,10 @@
 package fs
 
 import (
+	"fmt"
+	"strconv"
+	"strings"
+
 	"github.com/docker/libcontainer/cgroups"
 	"github.com/docker/libcontainer/configs"
 )
@@ -11,14 +15,25 @@ type HugetlbGroup struct {
 }
 
 func (s *HugetlbGroup) Apply(d *data) error {
-	// we just want to join this group even though we don't set anything
-	if _, err := d.join("hugetlb"); err != nil && !cgroups.IsNotFound(err) {
+	dir, err := d.join("hugetlb")
+	if err != nil && !cgroups.IsNotFound(err) {
 		return err
 	}
+
+	if err := s.Set(dir, d.c); err != nil {
+		return err
+	}
+
 	return nil
 }
 
 func (s *HugetlbGroup) Set(path string, cgroup *configs.Cgroup) error {
+	for _, hugetlb := range cgroup.HugetlbLimit {
+		if err := writeFile(path, strings.Join([]string{"hugetlb", hugetlb.Pagesize, "limit_in_bytes"}, "."), strconv.Itoa(hugetlb.Limit)); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


crosbymichael:
LGTM

crosbymichael:
Do you know of a better way to represent this data instead of something like `2M:2312323`?  Maybe a struct instead of parsing the string?

Mashimiao:
change it to struct is OK. But why do you think `2M:2312323` is not good?

Mashimiao:
ping @LK4D4

crosbymichael:
@Mashimiao just having type safety instead of string parsing in the API

Mashimiao:
@crosbymichael
modified and rebased

configs/hugetlb.go
```diff
@@ -0,0 +1,9 @@
+package configs
+
+type Hugetlb struct {
```
vmarmol: We should stick this one in the cgroups file since all the cgroup stuff is there

configs/hugetlb.go
```diff
@@ -0,0 +1,9 @@
...
+	// which type of hugepage to limit.
+	Pagesize string `json:"page_size"`
```
vmarmol: Isn't there a limited set of these? If so, what about an enum?

cgroups/stats.go
```diff
@@ -63,13 +63,24 @@ type BlkioStats struct {
...
+}
+
 type Stats struct {
-	CpuStats    CpuStats    `json:"cpu_stats,omitempty"`
-	MemoryStats MemoryStats `json:"memory_stats,omitempty"`
-	BlkioStats  BlkioStats  `json:"blkio_stats,omitempty"`
+	CpuStats     CpuStats                `json:"cpu_stats,omitempty"`
+	MemoryStats  MemoryStats             `json:"memory_stats,omitempty"`
+	BlkioStats   BlkioStats              `json:"blkio_stats,omitempty"`
+	HugetlbStats map[string]HugetlbStats `json:"hugetlb_stats,omitempty"`
```
vmarmol: Can we document what this is a map of please? I'm guessing hugepage size to stats about that size?

vmarmol:
Thanks for the PR @Mashimiao!

cgroups/stats.go
Mashimiao: Yes, you are right. I will document it.

configs/hugetlb.go
Mashimiao: I think the name `Hugetlb` makes you confused.
We already have HugetlbGroup in cgroup/fs/hugetlb.go
I will change the name of struct and file

configs/hugetlb.go
Mashimiao: No, I think it's better to dynamically get what hugepage the host support

Mashimiao:
ping @vmarmol

configs/hugetlb.go
vmarmol: Hmmm this type of stuff makes it a very hard API to consume. How would I know what values are allowed? How to I know which are available?

I'm OK starting with this, but can we at least put example of valid values? :)

configs/hugetlb.go
vmarmol: We should still put it in the cgroup file, the divisions of files there are very broad: cgroup, network, namespaces. This one is too fine to be in its own file.

cgroups/fs/hugetlb.go
```diff
@@ -25,5 +40,37 @@ func (s *HugetlbGroup) Remove(d *data) error {
 }
 
 func (s *HugetlbGroup) GetStats(path string, stats *cgroups.Stats) error {
+	pageSizes, err := getHugePageSize()
```
vmarmol: Let's only do the getHugePageSize() once in create and cache that.

Mashimiao: Fixed.

vmarmol: I mean to do this once in a `NewHugetlbGroup()` so we only need to do it once per run of libcontainer, not at every stats collection. Or do the hugepage sizes change frequently?

configs/hugetlb.go
vmarmol: ping on this

configs/hugetlb.go
vmarmol: ping on adding example.

cgroups/fs/hugetlb.go
Mashimiao: @vmarmol  Any XXXGroup is created When subsystems is created. And in Manager.Apply() used first. Do you mean put it into Manager.Apply()?

vmarmol: We can either make a New() and use that where we instantiate today: https://github.com/docker/libcontainer/blob/master/cgroups/fs/apply_raw.go#L26

Or we can probably do an init() function and save it as a global

Mashimiao:
@vmarmol  I have modified code. WDYT?

LK4D4:
LGTM
ping @vmarmol

vmarmol:
Sorry for the delay. LGTM

PR #576, opened 2015-05-08
Do not prevent mounts in /sys

crosbymichael:
Mounts in /sys like /sys/fs/cgroup are valid and should be allowed at
the libcontainer level.


jfrazelle:
LGTM

mrunalp:
LGTM

PR #577, opened 2015-05-10
Add exceptions for checkMountDest

hqhq:
We still should prevent mount in /sys for security, special
destionations like /sys/fs/cgroup we can treated as exceptions,
these can be unknown from users for now.

Signed-off-by: Qiang Huang h.huangqiang@huawei.com


crosbymichael:
ping @ewindisch can you provide some feedback here?

ewindisch:
:+1:  I absolutely prefer to whitelist than blacklist. If there are other paths we should whitelist, it would be good to get them in together before the next release.

LK4D4:
LGTM
Dunno what other paths though.
ping @crosbymichael

PR #578, opened 2015-05-11
Tighten proc and caps in apparmor policy

ewindisch:
Adds denies for various paths underneath /proc,
restricts capabilities. Following this patch,
adding capabilities beyond the default set will
not be possible without also specifying an alternative
AppArmor profile.

Signed-off-by: Eric Windisch <eric@windisch.us>


ewindisch:
May need more granular settings for /sys...

PR #579, opened 2015-05-11
AppArmor disable umount & deny proc files.

ewindisch:
If a container cannot mount, it cannot umount.

We should also restrict writing /proc/kcore and the
other paths libcontainer now masks.

This is a less aggressive prelude to #578.

Signed-off-by: Eric Windisch eric@windisch.us


LK4D4:
I personally know nothing about AppArmor.
but looks okay
LGTM
ping @crosbymichael

ewindisch:
Obsolete

PR #581, opened 2015-05-12
Allow restore to actually exec/create the libcontainer

boucher:
Right now, you can restore any image into a container, so long as that container has been started and then checkpointed. This patch lets you restore an image into a brand new container, without first requiring you to start and checkpoint it.


crosbymichael:
LGTM

PR #585, opened 2015-05-14
Cgroup Test: check whether this subsystem is supported

Mashimiao:

### DIFF
--- a/cgroups/fs/util_test.go
+++ b/cgroups/fs/util_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/docker/libcontainer/configs"
+	"github.com/docker/libcontainer/cgroups"
 )
 
 type cgroupTestUtil struct {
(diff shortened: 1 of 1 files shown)


hqhq:
Hi Mashimiao,

Unit test should not rely on the real environment, I tried to do this before, see https://github.com/docker/libcontainer/pull/433

Mashimiao:
@hqhq Yes, already got it.
I sent a PR to set cgroup net_cls, I found I can't get correct value from net_cls.classid.
I find that may because we are not in right cgroup dir.
So I Sent this PR.
But After sending this PR, I suddenly found we are testing in mock environment.
So, I changed my testcase and closed this PR.

hqhq:
Sure, maybe you can write a reason before closing PR, just saw you close and reopen and close, thought that all by mistake, it happens ;)

BTW, good work about cgroup enhancement, I always want to add all missing cgroup subsystems, just never got my hands free.

PR #586, opened 2015-05-14
container: don't use c.initProcess before it's set

avagin:
container.Checkpoint() doesn't clear c.initProcess and it's used on restore. A restored process is set from criuNotifications().

Cc: @boucher


boucher:
Verified this works.

PR #588, opened 2015-05-14
Update nsinit to use cwd as id

crosbymichael:
This includes another changes where by default it will use the current
cgroup as the parent for placing the container.  This makes things work
if a process supervisor is the parent of nsinit and it is setting up
    cgroups.

``` json
{
    "id": "jessie",
    "init_process_pid": 20587,
    "init_process_start": "15242763",
    "cgroup_paths": {
        "blkio": "/sys/fs/cgroup/blkio/user.slice/user-1000.slice/session-c2.scope/jessie",
        "cpu": "/sys/fs/cgroup/cpu/user.slice/user-1000.slice/session-c2.scope/jessie",
        "cpuacct": "/sys/fs/cgroup/cpuacct/user.slice/user-1000.slice/session-c2.scope/jessie",
        "cpuset": "/sys/fs/cgroup/cpuset/user.slice/user-1000.slice/session-c2.scope/jessie",
        "devices": "/sys/fs/cgroup/devices/user.slice/user-1000.slice/session-c2.scope/jessie",
        "freezer": "/sys/fs/cgroup/freezer/user.slice/user-1000.slice/session-c2.scope/jessie",
        "hugetlb": "/sys/fs/cgroup/hugetlb/user.slice/user-1000.slice/session-c2.scope/jessie",
        "memory": "/sys/fs/cgroup/memory/user.slice/user-1000.slice/session-c2.scope/jessie",
        "perf_event": "/sys/fs/cgroup/perf_event/user.slice/user-1000.slice/session-c2.scope/jessie"
    },
    "namespace_paths": {
        "NEWIPC": "/proc/20587/ns/ipc",
        "NEWNET": "/proc/20587/ns/net",
        "NEWNS": "/proc/20587/ns/mnt",
        "NEWPID": "/proc/20587/ns/pid",
        "NEWUSER": "/proc/20587/ns/user",
        "NEWUTS": "/proc/20587/ns/uts"
    }
}
```

Signed-off-by: Michael Crosby crosbymichael@gmail.com


nsinit/exec.go
```diff
@@ -51,7 +45,7 @@ func execAction(context *cli.Context) {
 	}
 	process := &libcontainer.Process{
 		Args:   context.Args(),
-		Env:    context.StringSlice("env"),
+		Env:    append(os.Environ(), context.StringSlice("env")...),
```
dqminh: i think this is somewhat unexpected as users now have to clear their env if they want to finely control the exec process's environment variable.

crosbymichael: Oh ya, this is for when you are running nsinit from a process supervisor the env comes from what you setup in a unit file or upstart script

mrunalp: Could we add a flag to toggle this behavior?

crosbymichael: Is there a reason why this would cause issues?

mrunalp: No issues really except that there won't be an easy way to exactly control the process's env.
So, if this is the default, maybe a --disable-env-inherit flag could retain the old behavior or --inherit-os-env to enable this.

LK4D4: I agree that this is little unexpected behavior by default. `--inherit-env` flag makes sense.

PR #592, opened 2015-05-15
cgroup memory: Enchance stats support of memory

Mashimiao:

### DIFF
--- a/cgroups/fs/memory.go
+++ b/cgroups/fs/memory.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/docker/libcontainer/cgroups"
 	"github.com/docker/libcontainer/configs"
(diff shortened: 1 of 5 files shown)


cgroups/stats.go
```diff
@@ -30,18 +30,21 @@ type CpuStats struct {
 	ThrottlingData ThrottlingData `json:"throttling_data,omitempty"`
 }
 
+type MemoryData struct {
+	Usage    uint64 `json:"usage,omitempty"`
+	MaxUsage uint64 `json:"max_usage,omitempty"`
+	Failcnt  uint64 `json:"failcnt"`
+}
 type MemoryStats struct {
```
vmarmol: This will be a breaking API change in Docker (since it uses the raw structures no?) @crosbymichael

cgroups/stats.go
```diff
@@ -30,18 +30,21 @@ type CpuStats struct {
...
 	// memory used for cache
 	Cache uint64 `json:"cache,omitempty"`
-	// maximum usage ever recorded.
-	MaxUsage uint64 `json:"max_usage,omitempty"`
-	// TODO(vishh): Export these as stronger types.
-	// all the stats exported via memory.stat.
-	Stats map[string]uint64 `json:"stats,omitempty"`
-	// number of times memory usage hits limits.
-	Failcnt uint64 `json:"failcnt"`
+	// usage of memory
```
vmarmol: WDYT: 
MemoryUsage -> Usage
MemswUsage -> SwapUsage
KmemUsage -> KernelUsage

cgroups/fs/memory.go
```diff
@@ -87,24 +87,86 @@ func (s *MemoryGroup) GetStats(path string, stats *cgroups.Stats) error {
...
+		return cgroups.MemoryData{}, fmt.Errorf("failed to parse memory.failcnt - %v", err)
 	}
-	stats.MemoryStats.Failcnt = value
+	memoryUsage.Failcnt = value
 
-	return nil
+	return memoryUsage, nil
+}
+
+func getMemswUsage(path string) (cgroups.MemoryData, error) {
```
vmarmol: we can probably merge these, the only difference is the names of the files no?

cgroups/stats.go
Mashimiao: Fine. Fixed

cgroups/fs/memory.go
Mashimiao: Fixed

cgroups/stats.go
Mashimiao: I think the change is necessary.
It can  describe the memory used more detailed

vmarmol: I agree that the change is preferable, but I don't know what API guarantees Docker makes in the stats API and I wanted to let @crosbymichael know :)

vmarmol:
LGTM, ping @rjnagal @mrunalp @crosbymichael for the second LGTM

cgroups/stats.go
crosbymichael: fine with me, we have independent structures for output of the API

rjnagal:
LGTM, new stats look fine.

cgroups/fs/memory.go
```diff
@@ -92,24 +93,53 @@ func (s *MemoryGroup) GetStats(path string, stats *cgroups.Stats) error {
...
-	stats.MemoryStats.MaxUsage = value
-	value, err = getCgroupParamUint(path, "memory.failcnt")
+	stats.MemoryStats.SwapUsage = swapUsage
+	kernelUsage, err := getMemoryData(path, "kmem")
 	if err != nil {
-		return fmt.Errorf("failed to parse memory.failcnt - %v", err)
+		return err
 	}
-	stats.MemoryStats.Failcnt = value
+	stats.MemoryStats.KernelUsage = kernelUsage
```
dqminh: should we ignore error when we cant get memory data from `memsw` and/or `kmem` as user kernel may not have them ?

Mashimiao: @dqminh
That's really a problem. I have changed my code, how do you think about it?

cgroups/fs/memory.go
```diff
@@ -94,24 +95,53 @@ func (s *MemoryGroup) GetStats(path string, stats *cgroups.Stats) error {
...
 	if err != nil {
-		return fmt.Errorf("failed to parse memory.usage_in_bytes - %v", err)
+		return err
 	}
-	stats.MemoryStats.Usage = value
-	stats.MemoryStats.Cache = stats.MemoryStats.Stats["cache"]
-	value, err = getCgroupParamUint(path, "memory.max_usage_in_bytes")
+	stats.MemoryStats.Usage = memoryUsage
+	swapUsage, err := getMemoryData(path, "memsw")
+	if err != nil && !strings.Contains(err.Error(), "no such file or directory") {
```
dqminh: This error check can be changed to `os.IsNotExist(err)`. Also I think its cleaner if we move that check inside `getMemoryData` so its look like:
```go
value, err := getCgroupParamUint(path, usage)
if err != nil {
  if os.IsNotExist(err) {
    return cgroups.MemoryData{}, nil
  }
  return cgroups.MemoryData{}, fmt.Errorf("...")
}
```
Then the consumer wont have to check for that error and just set the value to default value.

Wdyt ?

Mashimiao: I think it's fine. Fixed.

cgroups/fs/memory.go
```diff
@@ -94,24 +95,62 @@ func (s *MemoryGroup) GetStats(path string, stats *cgroups.Stats) error {
...
+	if name != "" {
+		moduleName = strings.Join([]string{"memory", name}, ".")
+	}
+	usage := strings.Join([]string{moduleName, "usage_in_bytes"}, ".")
+	maxUsage := strings.Join([]string{moduleName, "max_usage_in_bytes"}, ".")
+	failcnt := strings.Join([]string{moduleName, "failcnt"}, ".")
+
+	value, err := getCgroupParamUint(path, usage)
+	if err != nil {
+		if moduleName != "memory" && os.IsNotExist(err) {
```
dqminh: i think `moduleName` will never be `memory` here if name is not empty. It will be `memory.${name}`

Mashimiao: If name is not empty, of course moudleName will never be "memory".
If we want to get memory.usage_in_bytes, we will path empty as name, then moudleName will be "memory" here

dqminh: ahh, i got it now !

dqminh:
LGTM

dqminh:
@Mashimiao Thanks so much for your work !

PR #594, opened 2015-05-15
First attempt at an integration for checkpoint/restore.

boucher:
The test creates a "cat" process, checkpoints that process, then restores the process, making sure at each point that things are in the correct state.


boucher:
cc @avagin

avagin:
LGTM. Thanks.

boucher:
@crosbymichael Can you take a look?

crosbymichael:
thanks @boucher

PR #601, opened 2015-05-25
Add proc mount check for container's config

junxu:
In libcontainer, when we create a new container, it will read some files
in proc filesystem. So if the container config has configured NEWNS
namespace, it must also require the proc mount defined.

This pr add proc mount check in config validate before it really creates a container, 
We also add proc mount for sample configs, then nsinit won't fail to create containers
using these config

Signed-off-by: junxu xujun@cmss.chinamobile.com

### DIFF
--- a/configs/validate/config.go
+++ b/configs/validate/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/libcontainer/configs"
 )
@@ -33,6 +34,9 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 		return err
 	}
 	if err := v.usernamespace(config); err != nil {
+		return err
+	}
+	if err := v.procMount(config); err != nil {
 		return err
 	}
 	return nil
@@ -90,4 +94,17 @@ func (v *ConfigValidator) usernamespace(config *configs.Config) error {
 		}
 	}
 	return nil
+}
+
+func (v *ConfigValidator) procMount(config *configs.Config) error {
+	if !config.Namespaces.Contains(configs.NEWNS) {
+		return nil
+	}
+
+	for _, m := range config.Mounts {
+		if (strings.Trim(m.Source, " ") == "proc") && (strings.Trim(strings.Trim(m.Destination, " "), "/") == "proc") {
+			return nil
+		}
+	}
+	return fmt.Errorf("NEWNS namespace specified, but porc munt isn't enabled in the config")
 }
--- a/configs/validate/config_test.go
+++ b/configs/validate/config_test.go
@@ -1,0 +1,40 @@
+package validate
+
+import (
+	"testing"
+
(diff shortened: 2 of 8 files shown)


avagin:
I don't understand the explanation. If we are going to read files from /proc, we need to mount it. Otherwise we don't need to mount it. I don't see a reason to mount it each time. Could you elaborate?

junxu:
@avagin Thans for your comments. 

1. if a new container runs with a new mount namespace, it must mount the "proc" filesystem in "/proc".  a)the sample configs doesn't contain the “proc” in mounts， so when we use these configs to run a new container, it will fail.
b)So we add "proc" mount in these sample configs.

2. If create a container with a config which has setted the NEWNS namespace but without "proc" mount. the libcontainer will return error in very late stage of the container creation.
so we add the "proc" mount check for config validation in the early stage of the container creation, 
and return caller a missing "proc" mount message.


3. When we have checked the config missing the "proc" mount, there are two ways to resolve it.
a) Add the an default proc mount for config in code.
b) just return an error which will notify caller a missing "proc" mount message, and then the caller will known how to modify his config file. (We adopt in this PR)

avagin:
finalizeNamespace():
* opens /proc/self/fd to close file descriptors
* opens /proc/pid/status to get capabilities

I think we can rework code to work without proc, but I don't sure whether it's really needed.

avagin:
I don't like the idea to read data from container's /proc/. An user can overmount files there and show a wrong information. For example, the user can overmount /proc/pid/status and show an empty capability bounding set there. In this case libcontainer will not drop bits from it and the case the entered process may get extra capabilities.
I think libcontainer should not use container's /proc. @crosbymichael @LK4D4 what do you think about this?

LK4D4:
I think @avagin is right. We can avoid using proc in `finalizeNamespace` and it will be less hacky.

avagin:
I was wrong about leaking capabilities. We read /proc/pid/status to get a capability bounding set, but we don't use it.
```
newCapWhitelist
  NewPid
    c.Load
      os.Open(/proc/%d/status")
      fmt.Sscanf(line[4:], "nd:  %08x%08x", &c.bounds[1], &c.bounds[0])
dropBoundingSet
  w.pid.Clear(capability.BOUNDS)
       c.bounds[0] = 0
       c.bounds[1] = 0
```

avagin:
There is one more place in gocapability. It reads /proc/sys/kernel/cap_last_cap to set CAP_SETPCAP. If we show a fake value in this file, an entered process can get extra capabilities in the bounding set.
```
                if (1<<uint(CAP_SETPCAP))&data[0].effective != 0 {
                        for i := Cap(0); i <= CAP_LAST_CAP; i++ {
                                if c.Get(BOUNDING, i) {
                                        continue
                                }
                                err = prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0)

```

LK4D4:
@avagin We can read `CAP_LAST_CAP` on host and pass it as env to init process and set it there from env without using value from proc.

PR #602, opened 2015-05-25
Fix race in stats Manager

runcom:
Came across https://gist.github.com/runcom/5791bd0befbe8f0c7d3b while testing stats in Docker, hope the fix is right or fixing it Docker side is enough https://github.com/docker/docker/pull/13444

Signed-off-by: Antonio Murdaca <me@runcom.ninja>

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -44,6 +44,7 @@ type subsystem interface {
 }
 
 type Manager struct {
+	mu      sync.Mutex
 	Cgroups *configs.Cgroup
 	Paths   map[string]string
 }
@@ -82,7 +83,6 @@ type data struct {
 }
 
 func (m *Manager) Apply(pid int) error {
-
 	if m.Cgroups == nil {
 		return nil
 	}
@@ -128,14 +128,25 @@ func (m *Manager) Apply(pid int) error {
 }
 
 func (m *Manager) Destroy() error {
-	return cgroups.RemovePaths(m.Paths)
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if err := cgroups.RemovePaths(m.Paths); err != nil {
+		return err
+	}
+	m.Paths = make(map[string]string)
+	return nil
 }
 
 func (m *Manager) GetPaths() map[string]string {
-	return m.Paths
+	m.mu.Lock()
+	paths := m.Paths
+	m.mu.Unlock()
+	return paths
 }
 
 func (m *Manager) GetStats() (*cgroups.Stats, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	stats := cgroups.NewStats()
 	for name, path := range m.Paths {
 		sys, ok := subsystems[name]
--- a/cgroups/systemd/apply_systemd.go
+++ b/cgroups/systemd/apply_systemd.go
@@ -20,6 +20,7 @@ import (
 )
 
 type Manager struct {
+	mu      sync.Mutex
 	Cgroups *configs.Cgroup
 	Paths   map[string]string
 }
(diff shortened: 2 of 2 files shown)


cgroups/fs/apply_raw.go
```diff
@@ -137,7 +138,10 @@ func (m *Manager) GetPaths() map[string]string {
 
 func (m *Manager) GetStats() (*cgroups.Stats, error) {
 	stats := cgroups.NewStats()
-	for name, path := range m.Paths {
+	m.Lock()
+	paths := m.Paths
+	m.Unlock()
+	for name, path := range paths {
```
dqminh: hmm the race is between `GetStats` and `RemovePaths` right ? I think we should also lock RemovePaths in that case.

Also should we lock for the entire duration of GetStats() ? With this lock a cgroup can still be removed in the middle of getting stats out and it can be an error in that case.

runcom: oh right, RemovePaths needs locking too.. I'll update

cgroups/fs/apply_raw.go
```diff
@@ -44,6 +44,7 @@ type subsystem interface {
 }
 
 type Manager struct {
+	sync.Mutex
```
dqminh: nit: I prefer to declare the mutex as `mu sync.Mutex` so Lock and Unlock doesnt belong to the Manager public API.

cgroups/fs/apply_raw.go
```diff
@@ -128,14 +128,23 @@ func (m *Manager) Apply(pid int) error {
...
+	defer m.Unlock()
+	if err := cgroups.RemovePaths(m.Paths); err != nil {
+		return err
+	}
+	return nil
 }
 
 func (m *Manager) GetPaths() map[string]string {
+	m.Lock()
+	defer m.Unlock()
```
dqminh: nit: i think we can remove defer here and just Unlock after getting the paths.

cgroups/fs/apply_raw.go
vmarmol: +1

cgroups/fs/apply_raw.go
runcom: Ok it was like this but then we said something about locking for the whole GetStats (maybe in Docker not here)

cgroups/fs/apply_raw.go
runcom: Sure will do

cgroups/fs/apply_raw.go
runcom: @dqminh should this be done even in `Destroy` above?

runcom:
@dqminh updated

cgroups/fs/apply_raw.go
```diff
@@ -128,14 +128,25 @@ func (m *Manager) Apply(pid int) error {
 }
 
 func (m *Manager) Destroy() error {
-	return cgroups.RemovePaths(m.Paths)
+	m.mu.Lock()
+	paths := m.Paths
+	m.mu.Unlock()
+	if err := cgroups.RemovePaths(paths); err != nil {
```
rjnagal: we can just 'return cgroup.RemovePaths(..)' as before.

runcom: done (I don't really know why I did this before :/)

calavera:
LGTM

cgroups/fs/apply_raw.go
```diff
@@ -128,14 +128,22 @@ func (m *Manager) Apply(pid int) error {
 }
 
 func (m *Manager) Destroy() error {
-	return cgroups.RemovePaths(m.Paths)
+	m.mu.Lock()
+	paths := m.Paths
+	m.mu.Unlock()
+	return cgroups.RemovePaths(paths)
```
dqminh: i actually prefer to lock for entire duration of Destroy and reset Paths in the end
 
```go
m.mu.Lock()
defer m.mu.Unlock()
err := cgroups.RemovePaths(paths)
if err != nil {
  return err
}
m.Paths = make(map[string]string)
return nil
```

So the race between Destroy and GetStats will be removed. `Stats` calls in the middle of `Destroy` will wait and not return partial data. Wdyt ?

runcom: Make sense

crosbymichael:
LGTM

PR #603, opened 2015-05-26
Rebased: Additional groups lookup

dqminh:
Original PR #559:
> This PR modifies AdditionalGroups to be a string array and the lookup to translate to group ids is performed in the container as discussed in docker/docker#10717
Signed-off-by: Mrunal Patel mrunalp@gmail.com

This carries #559, did a rebase and implement the group lookup by scanning through the group file only once.

@mrunalp Can you take a look if this looks good ? I cant open the PR back to the original because of the rebase.

### DIFF
--- a/configs/config.go
+++ b/configs/config.go
@@ -119,7 +119,7 @@ type Config struct {
 
 	// AdditionalGroups specifies the gids that should be added to supplementary groups
 	// in addition to those that the user belongs to.
-	AdditionalGroups []int `json:"additional_groups"`
+	AdditionalGroups []string `json:"additional_groups"`
 
 	// UidMappings is an array of User ID mappings for User Namespaces
 	UidMappings []IDMap `json:"uid_mappings"`
--- a/user/user.go
+++ b/user/user.go
@@ -349,51 +349,59 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 	return user, nil
 }
 
-// GetAdditionalGroupsPath is a wrapper for GetAdditionalGroups. It reads data from the
-// given file path and uses that data as the arguments to GetAdditionalGroups.
+// GetAdditionalGroupsPath looks up a list of groups by name or group id
+// against the group file. If a group name cannot be found, an error will be
+// returned. If a group id cannot be found, it will be returned as-is.
 func GetAdditionalGroupsPath(additionalGroups []string, groupPath string) ([]int, error) {
-	var groupIds []int
-
-	for _, ag := range additionalGroups {
-		groupReader, err := os.Open(groupPath)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to open group file: %v", err)
-		}
-		defer groupReader.Close()
-
-		groupId, err := GetAdditionalGroup(ag, groupReader)
-		if err != nil {
-			return nil, err
-		}
-		groupIds = append(groupIds, groupId)
+	groupReader, err := os.Open(groupPath)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open group file: %v", err)
 	}
+	defer groupReader.Close()
 
-	return groupIds, nil
-}
-
-// GetAdditionalGroup looks up the specified group in the passed groupReader.
-func GetAdditionalGroup(additionalGroup string, groupReader io.Reader) (int, error) {
 	groups, err := ParseGroupFilter(groupReader, func(g Group) bool {
-		return g.Name == additionalGroup || strconv.Itoa(g.Gid) == additionalGroup
+		for _, ag := range additionalGroups {
+			if g.Name == ag || strconv.Itoa(g.Gid) == ag {
+				return true
+			}
+		}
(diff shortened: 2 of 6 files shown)


mrunalp:
@dqminh Thanks a lot for doing this. I tested it and it works fine :)

mrunalp:
@LK4D4 PTAL

user/user.go
```diff
@@ -348,3 +348,67 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+			}
+			// only add the group if we dont have it in the gid map values
+			var found bool
+			for _, gid := range gidMap {
+				if addGroup == gid {
+					found = true
+					break
+				}
+			}
+			if !found {
```
LK4D4: Here found can't be true, because you're break from loop when set true. So you don't need this variable at all.

user/user.go
```diff
@@ -348,3 +348,67 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open group file: %v", err)
+	}
+	defer groupReader.Close()
+
+	groups, err := ParseGroupFilter(groupReader, func(g Group) bool {
+		var found bool
+		for _, ag := range additionalGroups {
+			if g.Name == ag || strconv.Itoa(g.Gid) == ag {
+				found = true
```
LK4D4: same here, you can just return true here and false below.

user/user.go
```diff
@@ -348,3 +348,67 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+			}
+			// Ensure gid is inside gid range.
+			if addGroup < minId || addGroup > maxId {
+				return nil, ErrRange
+			}
+			// only add the group if we dont have it in the gid map values
+			var found bool
+			for _, gid := range gidMap {
+				if addGroup == gid {
+					found = true
```
LK4D4: Are you sure that you need break here and not continue?

user/user.go
```diff
@@ -348,3 +348,67 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 
 	return user, nil
 }
+
+// GetAdditionalGroupsPath looks up a list of groups by name or group id
+// against the group file. If a group name cannot be found, an error will be
+// returned. If a group id cannot be found, it will be returned as-is.
+func GetAdditionalGroupsPath(additionalGroups []string, groupPath string) (gids []int, err error) {
```
LK4D4: Not sure that you really used named returns here.

dqminh: i just used named return for some form of documentation. I can remove it if you want.

dqminh: actually i think i will just remove it to be consistent with the rest of the code.

user/user.go
dqminh: true, I will make the change.

user/user.go
dqminh: I think you are right. The logic here can be reworked to make it clearer.

user/user.go
```diff
@@ -348,3 +348,60 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+			if err != nil {
+				return nil, fmt.Errorf("Unable to find group %s", ag)
+			}
+			// Ensure gid is inside gid range.
+			if gid < minId || gid > maxId {
+				return nil, ErrRange
+			}
+			gidMap[gid] = struct{}{}
+		}
+	}
```
dqminh: @LK4D4 I reworked the loop so it's clearer now that we either look at the matched groups or save the new group id. PTAL.

LK4D4:
LGTM

LK4D4:
ping @crosbymichael

mrunalp:
@crosbymichael @vmarmol PTAL

configs/config.go
```diff
@@ -85,7 +85,7 @@ type Config struct {
 
 	// AdditionalGroups specifies the gids that should be added to supplementary groups
 	// in addition to those that the user belongs to.
-	AdditionalGroups []int `json:"additional_groups"`
```
vmarmol: Any reason for the change from int to string for groups?

vmarmol:
This looks good, but needs a rebase.

dqminh:
Going to rebase asap

configs/config.go
dqminh: i think it's because Docker wants to specify group as either gid or name.

dqminh:
@vmarmol rebased. PTAL

configs/config.go
tianon: Because we can for user (user:group syntax), so consistency :+1:

vmarmol: Thanks for explaining :)

vmarmol:
LGTM

cyphar:
Why wasn't I ping'd on this? This API isn't right, we need an `io.Reader` function so the tests are less dodgy.

PR #607, opened 2015-05-30
bug fix: slice bounds out of range

WeiZhang555:
If TestGenerateName() accept size parameter larger than 64, it will panic.
This patch fix this potential panic bug.

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>


LK4D4:
LGTM

PR #608, opened 2015-05-31
spec: Fix errors in file system mount points table.

davexunit:
I was following `SPEC.md` while setting up mount points for a container manually and noticed a couple of things that had to be changed in order for the `mount` calls to succeed.  "shm" is not a valid file system type, and there must be an "=" between "gid" and "5" in the devpts options.  I experienced these issues on Linux 4.0.4.


mrunalp:
LGTM

PR #609, opened 2015-06-01
A container can join pid namespace of another container.

### LINKED ISSUE #604: Share pid namespace between containers
As my approach in https://github.com/docker/docker/pull/13453, got a problem that libcontainer are not support setns to pid namespace properly, if we just setns to pid namespace, the process called setns is still outside of pid namespace.

For starting container, Docker use `nsexec.c` to work around this, but this problem still exists in `docker exec` for now, post this and hope we can discuss a solution for this.

dqminh:
Fix #604 

This supports setting `configs.NEWPID` to an existing PID namespace for a container's init process. It leverages the same C code infrastructure for `execin` to clone a new init process with the wanted PID namespace.

Edit: Because of some sharing in code between init and execin, execin also supports joining userns of the init process now. Init process cant join mnt/userns of another init process yet.

### DIFF
--- a/container_linux.go
+++ b/container_linux.go
@@ -128,7 +128,7 @@ func (c *linuxContainer) newParentProcess(p *Process, doInit bool) (parentProces
 		return nil, newSystemError(err)
 	}
 	if !doInit {
-		return c.newSetnsProcess(p, cmd, parentPipe, childPipe), nil
+		return c.newSetnsProcess(p, cmd, parentPipe, childPipe)
 	}
 	return c.newInitProcess(p, cmd, parentPipe, childPipe)
 }
@@ -157,7 +157,6 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 }
 
 func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) (*initProcess, error) {
-	t := "_LIBCONTAINER_INITTYPE=standard"
 	cloneFlags := c.config.Namespaces.CloneFlags()
 	if cloneFlags&syscall.CLONE_NEWUSER != 0 {
 		if err := c.addUidGidMappings(cmd.SysProcAttr); err != nil {
@@ -169,21 +168,60 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
 			cmd.SysProcAttr.Credential = &syscall.Credential{}
 		}
 	}
-	cmd.Env = append(cmd.Env, t)
 	cmd.SysProcAttr.Cloneflags = cloneFlags
+
+	// set init process environment
+	env := []string{"_LIBCONTAINER_INITTYPE=standard"}
+	var joinNamespaces configs.Namespaces
+	var doClone bool
+	nsMaps := make(map[configs.NamespaceType]string)
+	for _, ns := range c.config.Namespaces {
+		if ns.Path != "" {
+			nsMaps[ns.Type] = ns.Path
+			joinNamespaces = append(joinNamespaces, ns)
+			if ns.Type == configs.NEWPID {
+				doClone = true
+			}
+		}
+	}
+	if len(nsMaps) > 0 {
+		nsPaths, err := orderNamespacePaths(nsMaps)
+		if err != nil {
+			return nil, err
+		}
+		env = append(env, fmt.Sprintf("_LIBCONTAINER_NSPATH=%s",
+			strings.Join(nsPaths, ",")))
+	}
+	if doClone {
+		env = append(env, "_LIBCONTAINER_DOCLONE=true")
+	}
+	cmd.Env = append(cmd.Env, env...)
+
 	return &initProcess{
-		cmd:        cmd,
-		childPipe:  childPipe,
-		parentPipe: parentPipe,
-		manager:    c.cgroupManager,
-		config:     c.newInitConfig(p),
+		cmd:            cmd,
+		childPipe:      childPipe,
+		parentPipe:     parentPipe,
+		manager:        c.cgroupManager,
+		config:         c.newInitConfig(p),
+		joinNamespaces: joinNamespaces,
+		doClone:        doClone,
 	}, nil
 }
 
-func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) *setnsProcess {
+func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) (*setnsProcess, error) {
+	state, err := c.currentState()
+	if err != nil {
+		return nil, newSystemError(err)
+	}
+	nsPaths, err := orderNamespacePaths(state.NamespacePaths)
+	if err != nil {
+		return nil, newSystemError(err)
+	}
 	cmd.Env = append(cmd.Env,
-		fmt.Sprintf("_LIBCONTAINER_INITPID=%d", c.initProcess.pid()),
+		fmt.Sprintf("_LIBCONTAINER_NSPATH=%s", strings.Join(nsPaths, ",")),
 		"_LIBCONTAINER_INITTYPE=setns",
+		"_LIBCONTAINER_DOCLONE=true",
+		"_LIBCONTAINER_SETSID=true",
 	)
 	if p.consolePath != "" {
 		cmd.Env = append(cmd.Env, "_LIBCONTAINER_CONSOLE_PATH="+p.consolePath)
@@ -195,7 +233,7 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, 
 		childPipe:   childPipe,
 		parentPipe:  parentPipe,
 		config:      c.newInitConfig(p),
-	}
+	}, nil
 }
 
 func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
@@ -776,4 +814,29 @@ func (c *linuxContainer) currentState() (*State, error) {
 		}
 	}
 	return state, nil
+}
+
+// orderNamespacePaths sorts that namespace paths into a list of paths that we
+// can safely setns to.
+func orderNamespacePaths(namespaces map[configs.NamespaceType]string) ([]string, error) {
+	paths := []string{}
+	for _, nsType := range []configs.NamespaceType{
+		// TODO: enable configs.NEWUSER,
+		configs.NEWIPC,
+		configs.NEWUTS,
+		configs.NEWNET,
+		configs.NEWPID,
+		// mnt namespace must be the last one. After switching to mnt
+		// namespace, the old /proc will be inaccessible.
+		configs.NEWNS,
+	} {
+		if p, ok := namespaces[nsType]; ok && p != "" {
+			// only set to join this namespace if it exists
+			if _, err := os.Lstat(p); err != nil {
+				return nil, newSystemError(err)
+			}
+			paths = append(paths, p)
+		}
+	}
+	return paths, nil
 }
--- a/container_linux.go
+++ b/container_linux.go
@@ -821,14 +821,14 @@ func (c *linuxContainer) currentState() (*State, error) {
 func orderNamespacePaths(namespaces map[configs.NamespaceType]string) ([]string, error) {
 	paths := []string{}
 	for _, nsType := range []configs.NamespaceType{
-		// TODO: enable configs.NEWUSER,
 		configs.NEWIPC,
 		configs.NEWUTS,
 		configs.NEWNET,
 		configs.NEWPID,
-		// mnt namespace must be the last one. After switching to mnt
-		// namespace, the old /proc will be inaccessible.
 		configs.NEWNS,
+		// user namespace must be the last one because we need enough privilege
+		// to setns
+		configs.NEWUSER,
 	} {
 		if p, ok := namespaces[nsType]; ok && p != "" {
 			// check if the requested namespace is supported
--- a/container_linux.go
+++ b/container_linux.go
@@ -185,7 +185,7 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
 		}
 	}
 	if len(nsMaps) > 0 {
-		nsPaths, err := orderNamespacePaths(nsMaps)
+		nsPaths, err := c.orderNamespacePaths(nsMaps, true)
 		if err != nil {
 			return nil, err
 		}
@@ -213,7 +213,7 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, 
 	if err != nil {
 		return nil, newSystemError(err)
 	}
-	nsPaths, err := orderNamespacePaths(state.NamespacePaths)
+	nsPaths, err := c.orderNamespacePaths(state.NamespacePaths, false)
 	if err != nil {
 		return nil, newSystemError(err)
 	}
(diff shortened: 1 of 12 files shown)


nsenter/nsexec.c
```diff
@@ -60,6 +60,62 @@ static int clone_parent(jmp_buf * env)
 	return child;
 }
 
+void nsinit_pid()
```
mrunalp: Could we combine this logic into nsexec() itself?

dqminh: i tried that but the end result wasnt as clean. Since spawning an init process with a known pid namespace and exec a new process inside another container are two very different operations , i think that having two functions describes the intent better.

avagin:
The same hack is required for mount and user namespaces, because a multithreaded process can't change them.
http://man7.org/linux/man-pages/man2/setns.2.html
'''
A process may not be reassociated with a new mount namespace if it is multithreaded.
'''

What do you think about creating a directory with namespaces in container's root directory. We can bind-mount all required namespaces there before starting the container and then bind-mount container's namespace files to use them for executing processes inside CT. In this case we can use the same code of nsexec.c for starting the first and other processes.

LK4D4:
I agree with @avagin that bind-mounting namespace can be right way to do such stuff, because `nsexec.c` supposed to be very small layer only for setting namespaces and expanding it can lead us to support pretty big C code in golang runtime, which probably will hit some limitations.
I'm not against merging this though if this feature is needed by someone right now :)

dqminh:
@LK4D4 There's a few related issues in docker https://github.com/docker/docker/pull/13453 and https://github.com/docker/docker/issues/10163. Overall i think it's better to do this properly (if this is not the best way) since 1.7 is already in RC now so we have time :)

@avagin I dont understand the part about bind-mounting namespaces. Do you have an example for that ? My current understanding of that is that we have a special directory that contains namespace descriptors that we can detect when starting the container and use them to setns with some code from nsexec.c ?

mrunalp:
@dqminh The suggestion is to bind mount /proc/<pid>/ns/* files into some directory for each container (Just like how ip netns add works. It bind mounts net namespace fds under /var/run/netns). After that is done, new processes could use those paths to join the container. This could also be used for joining subset of other container's namespaces.

@avagin could correct me if I am wrong.

mrunalp:
@dqminh Example https://gist.github.com/mrunalp/5290c92b32fe709aadd3

LK4D4:
Also, you can destroy original namespace, so bind-mounting is like "copy" namespace.

mrunalp:
Right, as long the mounted namespace fds exists, the original container can just die without affecting those that joined those namespace fds.

mrunalp:
Not sure how that will behave will pid namespaces though since pid 1 going down brings down the other processes.

dqminh:
> The suggestion is to bind mount /proc/self/ns/* files into some directory for each container (Just like how ip netns add works

Ah i see. But that would not change the current implementation right, since all libcontainer needs to know is still the paths to the namespace descriptors.

> Not sure how that will behave will pid namespaces though since pid 1 going down brings down the other processes.

i think pid 1 going down will kill the rest of the processes, and you cant join using the open file descriptor anymore (http://man7.org/linux/man-pages/man7/pid_namespaces.7.html).

avagin:
> Ah i see. But that would not change the current implementation right, since all libcontainer needs to know is still the paths to the namespace descriptors.
@dqminh I don't understand what you mean. Could you elaborate?

LK4D4:
@dqminh for main process we use Cloneflags for creating namespaces. New idea is to create namespaces separately and then pass them even for first process.

avagin:
@LK4D4 @dqminh I don't suggest to create namespaces separately. For example, It's impossible for pidns. For the init process we will mount all existent namespaces there and set clone flags for others.

LK4D4:
@avagin What point of bind-mounting then? Just to collect them in one directory?
I think the we can pass just space-separated namespace list to nsexec.c where it will be generated from `config.Namespaces` for init and from `state.NamespacePaths` for other.

dqminh:
Right, we cant really seperate the creation of namespaces and processes at least for pid namespace. It always requires an anchor process.

The current way joining namespaces works for init process is that container has `config.Namespaces` which keys values mapping of namespace to their descriptor path. The value can either be empty ( which then the key is used in CloneFlags ), or points to a valid namespace descriptor ( which we use to join namespace in `joinExistingNamespaces` ). So from libcontainer, the only requirements for init to join different namespaces is to set the key-value pair in `config.Namespaces`

This works for net ( the only current usecase afaik ), but doesnt work for pid ( and probably mount/userns as @avagin pointed out ). This patch adds another intermediate clone to put the init process in correct pid namespaces.

My understanding is that even if we bind-mount namespaces, we will still pass the mount destination as value to `config.Namespaces` and perform `setns` before exec-ing the target process, so it will not be different from the current way.

> I think then we can pass just space-separated namespace list to nsexec.c where it will be generated from config.Namespaces for init and from state.NamespacePaths for other.

Hmm yes, that is another way to do it so we can reuse that blocks of code between init and execin. It also helps when we want to join another namespaces besides pid. Probably joinExistingNamespaces can be removed in that case.

avagin:
> @avagin What point of bind-mounting then? Just to collect them in one
directory?

* collect in one directory
* be independent from the init process. If we have a container without
pidns, we will able to enter into it even when the first process died. For
container with pidns, this approach avoids a problem when the init process
died and another process is created with the same pid.

I'm not sure that we need to do this. It's just an idea how it can be fixed.

> I think the we can pass just space-separated namespace list to nsexec.c
where it will be generated from config.Namespaces for init and from
state.NamespacePaths for other.

Yes, we can.

dqminh:
@LK4D4 @avagin i added some code to pass a list of namespaces to be joined to both init process and execin process. So they behave the same now on `setns` part.

All `setns` now is done in C layer (`joinExistingNamespaces` is not necessary)

nsenter/nsexec.c
```diff
@@ -103,46 +95,37 @@ void nsexec()
...
-		if (fstatat(tfd, namespaces[i], &st, AT_SYMLINK_NOFOLLOW) == -1) {
-			// Ignore nonexistent namespaces.
-			if (errno == ENOENT)
-				continue;
-		}
-
-		fd = openat(tfd, namespaces[i], O_RDONLY);
+	char *ns, *saveptr;
+	while ((ns = strtok_r(nspaths, ",", &saveptr))) {
+		fd = open(ns, O_RDONLY);
```
avagin: If you enter into a mount namespace, you can't be sure that you will able to enter in other namespaces, because nspaths may be inaccessible from this mount namespace.

container_linux.go
```diff
@@ -169,21 +168,58 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
...
+	var joinNamespaces configs.Namespaces
+	var nsPaths []string
+	var doClone bool
+	for _, ns := range c.config.Namespaces {
+		if ns.Path != "" {
+			joinNamespaces = append(joinNamespaces, ns)
+			nsPaths = append(nsPaths, ns.Path)
+			if ns.Type == configs.NEWPID {
+				doClone = true
+			}
```
avagin, line 184: You can't to enter into namespaces in a random order. For example, when you enter into a userns, you may lose permissions to enter in other namespaces.

container_linux.go
```diff
@@ -169,21 +168,58 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
...
+		manager:        c.cgroupManager,
+		config:         c.newInitConfig(p),
+		joinNamespaces: joinNamespaces,
+		doClone:        doClone,
 	}, nil
 }
 
 func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) *setnsProcess {
+	var nsPaths []string
+	for _, ns := range []string{"ipc", "uts", "net", "pid", "mnt"} {
```
avagin: Could you add "user" to this array?

container_linux.go
```diff
@@ -169,21 +168,58 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
...
+		path := fmt.Sprintf("/proc/%d/ns/%s", c.initProcess.pid(), ns)
```
avagin: The init process can died, but this code will not notice this.

nsenter/nsexec.c
dqminh: right, I think the order of nspaths has to be set from the caller ( which we are not doing now for init, but we do for exec ). wdyt ?

container_linux.go
dqminh: Can I add that in the next patch ? I want to keep the behavior the same as now.

dqminh: actually we probably can use `state.NamespacePaths` here instead of getting the proc path. We still need some code to sort that namespace paths so it enters by order, that code can be shared with the init process above.

container_linux.go
dqminh: right, i think in the old way we check for the existence of `/proc/pid/ns` to determine if the init is alive. I guess we can do the same here.

avagin: In the old way we open("/proc/pid/ns") and use fstatat(fd, ns).
We call stat to determine whether a namespace is supported by kernel. We can do lstat("/proc/self/%s", ns) for this.

nsenter/nsexec.c
avagin: I agree that a caller can do this.

process_linux.go
```diff
@@ -226,7 +266,14 @@ func (p *initProcess) wait() (*os.ProcessState, error) {
...
 	// we should kill all processes in cgroup when init is died if we use host PID namespace
-	if p.cmd.SysProcAttr.Cloneflags&syscall.CLONE_NEWPID == 0 {
+	// If the container joins another PID namespace, do not do anything.
+	var joinPidns bool
+	for _, ns := range p.joinNamespaces {
+		if ns.Type == configs.NEWPID {
+			joinPidns = true
+		}
+	}
+	if !joinPidns && p.cmd.SysProcAttr.Cloneflags&syscall.CLONE_NEWPID == 0 {
```
avagin: CLONE_NEWPID is set only when a process doesn't join another pidns, isn't it?

nsenter/nsenter_test.go
```diff
@@ -1,91 +0,0 @@
...
-	"os/exec"
-	"strings"
-	"testing"
-)
-
-type pid struct {
-	Pid int `json:"Pid"`
-}
-
-func TestNsenterAlivePid(t *testing.T) {
```
avagin, line 16: Why do you think these tests are useless?

process_linux.go
dqminh: Yes. Actually i think this code is not needed, since it should be safe to kill the current container's processes because two containers dont share cgroups. I will do some test and remove this if it's the case. Thanks for noticing it !

nsenter/nsenter_test.go
dqminh, line 16: Sorry i forgot to restore it when i changed the env. It was deleted because it has to be rewritten to test for invalid namespaces path, not invalid pids anymore since we only care about the namespace paths now.

I will restore it with new environment variables.

nsenter/nsexec.c
dqminh: done. The order of namespaces is now enforced in `orderNamespacePaths`. I also tried to add NEWUSER there but some tests failed. Added a note on NEWUSER and will try again in another patch only for NEWUSER.

container_linux.go
dqminh: in `orderNamespacePaths`, we check for Lstat of the namespaces and fail if it doesnt exist.

process_linux.go
dqminh: done. The error I noticed was because two container in tests share the same cgroups. Changed the tests to use different cgroups for the containers.

nsenter/nsenter_test.go
dqminh, line 16: done. I restored nsenter_test with 2 tests for valid and invalid namespaces.

dqminh:
@avagin I pushed a new patch that unifies the ordering of namespaces in init and exec, and a few more changes. PTAL.

Also i notice that joining mount namespace for init process will not work right now because it's still trying to setupRootfs. I can write a new PR to fix that ( if we want a init process to join mnt namespace of another container ? )

container_linux.go
```diff
@@ -169,21 +168,60 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
...
+			joinNamespaces = append(joinNamespaces, ns)
+			if ns.Type == configs.NEWPID {
+				doClone = true
+			}
+		}
+	}
+	if len(nsMaps) > 0 {
+		nsPaths, err := orderNamespacePaths(nsMaps)
+		if err != nil {
+			return nil, newSystemError(err)
```
avagin: orderNamepsapcePaths() returns SystemError

container_linux.go
```diff
@@ -777,3 +815,37 @@ func (c *linuxContainer) currentState() (*State, error) {
...
+func orderNamespacePaths(namespaces map[configs.NamespaceType]string) ([]string, error) {
+	paths := []string{}
+	for _, nsType := range []configs.NamespaceType{
+		// TODO: enable configs.NEWUSER,
+		configs.NEWIPC,
+		configs.NEWUTS,
+		configs.NEWNET,
+		configs.NEWPID,
+		// mnt namespace must be the last one. After switching to mnt
+		// namespace, the old /proc will be inaccessible.
```
avagin: I think USERNS must be the last one;). We will need to open a file descriptor for it before switching the mount namespace.

container_linux.go
dqminh: done

container_linux.go
dqminh: hmm yes i think so. WDYT about opening all file descriptors first before `setns` so it's consistent. I think doing that helps with ordering too (i.e., order doesnt matter). Something like:

```
fds = []int{}
for ns in namespaces { fds = append(fds, open(ns) }
for fd in fds { setns(fd, 0) }
```

avagin: I think it's ok. But userns must be the last one, otherwise you will not have permissions to switch other namespaces.
https://github.com/torvalds/linux/blob/master/fs/namespace.c#L3235

dqminh: added. Joining user namespace will only be applicable for execin process for now. There are some trouble with EPERM when setting up mnt/net when i was trying to let an init process join another init process in test. I didnt debug much further.

I figure that joining pid namespace is the most obvious usecase now. I can look into other usecase after (mnt/userns) if they make sense.

container_linux.go
```diff
@@ -777,3 +815,40 @@ func (c *linuxContainer) currentState() (*State, error) {
...
+	paths := []string{}
+	nsTypes := []configs.NamespaceType{
+		configs.NEWIPC,
+		configs.NEWUTS,
+		configs.NEWNET,
+		configs.NEWPID,
+		configs.NEWNS,
+	}
+	// For now, only join user namespace if this is an exec in process and the
+	// container supports user namespace
```
avagin, line 831: I would like to have explanation about this restriction.

nsenter/nsexec.c
```diff
@@ -105,47 +110,64 @@ void nsexec()
...
 			exit(1);
 		}
-		// Set the namespace.
-		if (setns(fd, 0) == -1) {
-			pr_perror("Failed to setns for %s", namespaces[i]);
+		nsList[i] = ns;
+		nspaths = NULL;
+	}
+	for (i = 0; i < nsLen; i++) {
+		if (fds[i] != -1 && setns(fds[i], 0) != 0) {
```
avagin, line 143: fds[i] can't be negative here, cat it be?

avagin:
LGTM (except a few minor comments).

hqhq:
@dqminh Can you backport this PR to runc? Thanks.

dqminh:
@hqhq Thanks for reminding me !. Things are a bit hectic on my side this week, but i should be able get it done by the weekend (or sooner if possible)

LK4D4:
ported to runc

PR #610, opened 2015-06-02
linux_container: fix error handling in start

Snorch:
if parent.terminate goes OK we will never know about parent.start error

Signed-off-by: Pavel Tikhomirov <snorcht@gmail.com>


LK4D4:
I don't get it. `err` is local to if block, so all should be ok.
http://play.golang.org/p/xAgwIR9k4q
Maybe I just didn't understand idea of fix.

Snorch:
Thank you for your polite answer, sorry, I missed that.

PR #614, opened 2015-06-04
Fix hack/validate.sh

LK4D4:
Was broken because of update in docker
ping @crosbymichael @jfrazelle


rjnagal:
LGTM

MAKEDIR change was a while back. Has it been broken since then?

LK4D4:
I think yes. Now we see this in Michael's seccomp PR.

vmarmol:
LGTM

PR #618, opened 2015-06-08
linux: Convert dup2 calls to dup3

glevand:
Convert syscall.Dup2 calls to syscall.Dup3.  The dup2 syscall is depreciated
and is not available on some architectures.  Fixes build errors like these when
building for arm64:

  console_linux.go: undefined: syscall.Dup2

Signed-off-by: Geoff Levand <geoff@infradead.org>

### DIFF
--- a/console_linux.go
+++ b/console_linux.go
@@ -92,7 +92,7 @@ func (c *linuxConsole) mount(rootfs, mountLabel string, uid, gid int) error {
 	return syscall.Mount(c.slavePath, dest, "bind", syscall.MS_BIND, "")
 }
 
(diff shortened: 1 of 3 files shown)


mrunalp:
LGTM

LK4D4:
@glevand Makes sense also change in `nsenter/nsexec.c`.

avagin:
@LK4D4 in nsexec.c dup2 is a glibc wrapper which handles the case when the dup2 syscall is absent. We can change nsexec.c to be consistant.

glevand:
Added a patch to cover nsenter.

LK4D4:
LGTM

PR #619, opened 2015-06-09
Fix relabel to allow volume mounting of /

rhatdan:
Current libcontainer in docker-1.7 will not allow you to volume mount /, /etc, and /usr into a container.  You even if you have not asked SElinux to relabel.  This patch will allow you to mount if you did not ask to relabel.

Needs to be cherrypicked into your docker-1.7 release.

### DIFF
--- a/label/label_selinux.go
+++ b/label/label_selinux.go
@@ -105,13 +105,13 @@ func Relabel(path string, fileLabel string, relabel string) error {
 	if fileLabel == "" {
 		return nil
 	}
+	if !strings.ContainsAny(relabel, "zZ") {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


rhatdan:
I just found this on docker-1.7 we were attempting to 

docker run --priviliged -v /:/host fedora /bin/sh

And it blows up saying that you can not relabel.

LK4D4:
LGTM
@rhatdan Do you want this to be included in 1.7?

rhatdan:
Yes, otherwise Systems with SELinux will break if you attempt ot mount one of these three directories into the container.

mrunalp:
LGTM

PR #622, opened 2015-06-11
Add the memory swappiness tuning support to libcontainer

ktraghavendra:
Signed-off-by: Raghavendra K T <raghavendra.kt@linux.vnet.ibm.com>

### DIFF
--- a/configs/cgroup.go
+++ b/configs/cgroup.go
@@ -87,6 +87,9 @@ type Cgroup struct {
 	// Whether to disable OOM Killer
 	OomKillDisable bool `json:"oom_kill_disable"`
 
+	// Tuning swappiness behaviour per cgroup
+	MemorySwappiness int64 `json:"memory_swappiness"`
+
 	// Set priority of network traffic for container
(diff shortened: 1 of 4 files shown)


dqminh:
LGTM

configs/cgroup.go
```diff
@@ -87,6 +87,9 @@ type Cgroup struct {
 	// Whether to disable OOM Killer
 	OomKillDisable bool `json:"oom_kill_disable"`
 
+	// Tuning swappiness behaviour per cgroup
+	MemSwappiness int64 `json:"swappiness"`
```
LK4D4: Can we call it `MemorySwappiness` for consistency? and move it to another `Memory..` fields?
Also as I see we don't use short names in json, so should be `memory_swappiness`

LK4D4:
I have naming nits, otherwise patch is great.

configs/cgroup.go
ktraghavendra: Thanks @dqminh @LK4D4 for review. Changing the naming and pushing to branch

LK4D4:
LGTM

PR #623, opened 2015-06-11
 Mount rootfs PRIVATE but volume mounts shared

rhatdan:
Currently we are remounting the entire OS as PRIVATE which breaks
shared volume mounts. This patch only mounts the rootfs as PRIVATE
which allows the volume mounts to be shared.

Docker-DCO-1.1-Signed-off-by: Dan Walsh dwalsh@redhat.com (github: rhatdan)

Docker-DCO-1.1-Signed-off-by: Dan Walsh dwalsh@redhat.com (github: rhatdan)

### DIFF
--- a/rootfs_linux.go
+++ b/rootfs_linux.go
@@ -168,6 +168,10 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 			if err := syscall.Mount("", dest, "none", uintptr(syscall.MS_PRIVATE), ""); err != nil {
 				return err
 			}
+		} else {
+			if err := syscall.Mount("", dest, "none", uintptr(syscall.MS_SLAVE), ""); err != nil {
+				return err
+			}
 		}
 	case "cgroup":
 		mounts, err := cgroups.GetCgroupMounts()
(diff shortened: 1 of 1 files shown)


rhatdan:
Does this solve the problem with the following 

https://github.com/docker/docker/pull/13854

crosbymichael:
Why would you not just set the MS_SLAVE flag in the mount's flags?

rhatdan:
@crosbymichael Won't work.  If you mount / as RPRIVATE all sharing is over FOREVER.  You can not get sharing back.  Mounting this way allows you to change the sharing per mount point.  

The question is does this fix the issues you were seeing?

We are working on another proposal to allow us to set this flag on a container by container basis.

https://github.com/docker/libcontainer/pull/632#issuecomment-112177648

rhatdan:
@crosbymichael Did you ever get a chance to look at this.

rootfs_linux.go
```diff
@@ -168,6 +168,10 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 			if err := syscall.Mount("", dest, "none", uintptr(syscall.MS_PRIVATE), ""); err != nil {
 				return err
 			}
+		} else {
+			if err := syscall.Mount("", dest, "none", uintptr(syscall.MS_SLAVE), ""); err != nil {
```
avagin, line 172: Could you explain this chunk?

rhatdan, line 172: If the mount was not specified as PRIVATE ,then we default to SLAVE.

avagin, line 172: Can I specify syscall.MS_SHARED? If yes, should we handle it?

rhatdan, line 172: Not yet, there is a separate pull request being worked on for handling all three requests.  

You really want that on a container by container basis.  The problem we have now is the entire host OS is being setup Private, which means no mount points can ever be made SHARED or SLAVE.  

This change is just to make ROOTFS PRIVATE allowing mountpoints outside of ROOTFS to be mounted as PRIVATE.  If this works and does not cause the problem @crosbymichael pulled the previous patch for, we can continue working on allowing users to specify differnet mount propogation rules on a container by container basis.

PR #624, opened 2015-06-11
Use simpler parsing of /proc/self/mountinfo for FindCgroupMountpoint

LK4D4:
It sped up this function x10 times(because of not using Ssprintf).
It was one of two major performance drawbacks in docker, because we're parsing
this file pretty often.


crosbymichael:
LGTM

mrunalp:
Nice! :)
LGTM

PR #631, opened 2015-06-15
init_linux: Inject the incoming UID (-u) into /etc/passwd

shaded-enmity:
This patch inserts the incoming `UID`/`GID` specified on the command line as:

`docker run -u 1234 image`

into `/etc/passwd` under the following circumstances:

1) The file already exists
2) `DockerUser` is not defined in the file
3) The incoming `UID` does not belong to any user as specified in the file

The rationale is that it's rather common idiom to use `getpwent()` and the like to get information about the `UID`, which fails if the user record is not present in the file, causing programs like `postgresql` to not work at all under a random `UID`.

Signed-off-by: Pavel Odvody podvody@redhat.com

### DIFF
--- a/init_linux.go
+++ b/init_linux.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -112,6 +113,12 @@ func finalizeNamespace(config *initConfig) error {
 	if err != nil {
 		return err
 	}
+
+	// inject /etc/passwd until we have sufficient privs to do that
+	if err := injectUserPasswd(config); err != nil {
+		return err
+	}
+
 	// drop capabilities in bounding set before changing user
 	if err := w.dropBoundingSet(); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


init_linux.go
```diff
@@ -138,6 +145,72 @@ func finalizeNamespace(config *initConfig) error {
...
+			return err
+		}
+
+		if len(found) == 0 {
+			/* No DockerUser in /etc/passwd - define one */
+			if fp, err := os.OpenFile("/etc/passwd", os.O_WRONLY|os.O_APPEND, os.FileMode(0666)); err == nil {
+				fp.WriteString(fmt.Sprintf("%s:x:%s:%s:Docker Mapped User:/:/sbin/nologin\n", "DockerUser", suid, sgid))
+				fp.Close()
+			} else {
+				return nil
```
cyphar: These chained `return nil`s aren't necessary. Just drop through to the end of the function. :)

shaded-enmity: Removed the `return nil` inception and rebased. Thanks for the comment :)

mrunalp:
The concern here is that we are modifying the rootfs and what happens on docker commit.

shaded-enmity:
@mrunalp Yes, I tried with putting this change to the topmost `ro` layer where `.dockerinit` lives, the problem is that at that time you can't read in-container `/etc/passwd` to figure out the overlaps.

How is `/etc/hosts` and `/etc/hostname` handled with respect to that?

shaded-enmity:
On a second thought I think I should put the injection right after all is mounted, and before `chroot` / `pivot`, that way I should be able to read the in-container `/etc/passwd` and also still be able to modify the upmost `ro` layer that doesn't get commited.

What do you think? @mrunalp

mrunalp:
@shaded-enmity I am not sure.  @LK4D4 @crosbymichael Any thoughts?

PR #634, opened 2015-06-16
gofmt to fix formatting

unclejack:


dqminh:
LGTM.

I think we should add gofmt to the list of things to check for on CI.

mrunalp:
LGTM

LK4D4:
@dqminh It is already there https://jenkins.dockerproject.org/job/Libcontainer-PRs/683/console:
```
Congratulations!  All Go source files are properly formatted.
```
But seems like not strict enough or I don't know.

mrunalp:
Yeah, not sure why this wasn't caught. It should have been by what we have.

rjnagal:
I think the CI script only checks format on changed files. It would not do anything for existing code.

PR #637, opened 2015-06-18
Fix kmem limit set

hqhq:
Currently we can't start container with kmem limit, because we
set kmem limit after processes joined to cgroup, we'll get device
busy error in this case.

Fix it by moving set before join.

### DIFF
--- a/cgroups/fs/memory.go
+++ b/cgroups/fs/memory.go
@@ -18,19 +18,28 @@ type MemoryGroup struct {
 }
 
 func (s *MemoryGroup) Apply(d *data) error {
-	dir, err := d.join("memory")
+	path, err := d.path("memory")
 	if err != nil && !cgroups.IsNotFound(err) {
+		return err
+	}
+	if err := os.MkdirAll(path, 0755); err != nil && !os.IsExist(err) {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


dqminh:
LGTM.

IIRC, kmem is the only thing that has this requirement. Please correct me if i'm wrong :D

hqhq:
@dqminh Yes, AFAIK only kmem has this requirement.

hqhq:
Sorry I forgot systemd part, updated now, please review again.

mrunalp:
@vmarmol @rjnagal PTAL

rjnagal:
yes, this is a kmem requirement as limiting can't happen when a cgroup already has a child:

from kernel docs:
"Kernel memory won't be accounted at all until limit on a group is set. This
allows for existing setups to continue working without disruption.  The limit
cannot be set if the cgroup have children, or if there are already tasks in the
cgroup. Attempting to set the limit under those conditions will return -EBUSY.
When use_hierarchy == 1 and a group is accounted, its children will
automatically be accounted regardless of their limit value."

Does it make sense to do Set before Join for everything?

rjnagal:
LGTM

PR #638, opened 2015-06-18
Avoid trying to access cpu.shares when it doesn't exist

lizf-os:
Even if cpu cgroup controller is enabled it's still possible that
cpu.shares doesn't exist.

This is the case when the kernel config has CONFIG_CGROUP_SCHED
enabled but CONFIG_FAIR_GROUP_SCHED disabled. Then docker fails to
start containers even --cpu-shares isn't specified.

$ sudo docker run -i -t ubuntu:14.04
Error response from daemon: Cannot start container 5600ae87eb9d0eca49f6bcee012247d6b4beb49c426d6cf17e2456279f9311f6: [8] System error: open /sys/fs/cgroup/cpu/docker/5600ae87eb9d0eca49f6bcee012247d6b4beb49c426d6cf17e2456279f9311f6/cpu.shares: no such file or directory

Signed-off-by: Zefan Li <lizefan@huawei.com>

### DIFF
--- a/cgroups/fs/apply_raw.go
+++ b/cgroups/fs/apply_raw.go
@@ -304,6 +304,10 @@ func removePath(p string, err error) error {
 func CheckCpushares(path string, c int64) error {
 	var cpuShares int64
 
+	if c == 0 {
+		return nil
+	}
+
 	fd, err := os.Open(filepath.Join(path, "cpu.shares"))
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


rjnagal:
What is the use-case for having CGROUP_SCHED on but shares disabled?
Does any distro do that?

lizf-os:
@rjnagal, I don't think any distro will do that, but there are various customized kernels out there. I don't know any real use-case, but enabling CFS_BANDWIDTH and/or RT_GROUP_SCHED while disabling FAIR_GROUP_SCHED doesn't sound very insane to me.

rjnagal:
The change itself looks fine to me. I was curious if disabling has any benefits.

LGTM

mrunalp:
LGTM

PR #642, opened 2015-06-19
Ensure all parent dirs are properly setup

crosbymichael:
Even if libcontainer does not create the directories for the cpuset
cgroup we should ensure that they are properly populated with the
parent's cpus and mems values.  Some systems create the directory
structures but do not correctly populate the values and causes our
implementation to fail.

The change in this code is that instead of stopping when we find the first parent directory in `cpuset`, we keep moving up the hierarchy until we hit the root of the `cpuset` cgroup to ensure that every parent is correctly created. 

### DIFF
--- a/cgroups/fs/cpuset.go
+++ b/cgroups/fs/cpuset.go
@@ -21,7 +21,6 @@ func (s *CpusetGroup) Apply(d *data) error {
 	if err != nil && !cgroups.IsNotFound(err) {
 		return err
 	}
-
 	return s.ApplyDir(dir, d.c, d.pid)
 }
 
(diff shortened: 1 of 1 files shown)


cgroups/fs/cpuset.go
```diff
@@ -87,19 +87,14 @@ func (s *CpusetGroup) getSubsystemSettings(parent string) (cpus []byte, mems []b
 // ensureParent ensures that the parent directory of current is created
```
LK4D4: Probably comment needs updating.

calavera:
LGTM

LK4D4:
LGTM

PR #644, opened 2015-06-20
New netlink library

LK4D4:
I want to play with `ipvlan` networking with libcontainer, therefore I need new netlink library. I replaced all current calls with it.


LK4D4:
ported to runc

PR #645, opened 2015-06-21
Do not generate/load AppArmor profiles

ewindisch:
Profiles for AppArmor should be specified by consuming
applications. A one-size-fits all policy specific to Docker
does not belong in libcontainer.

This change eliminates loading of the policy from libcontainer.
It also eliminates the check for the userland tools as these
were only needed for libcontainer's policy loader. Applying
profiles does not require the userland tools and some
systems may, in fact, have functional AppArmor without
having these tools on the host.

Signed-off-by: Eric Windisch <eric@windisch.us>


ewindisch:
Docker will need https://github.com/docker/docker/pull/13144 before updating libcontainer if this is applied.

LK4D4:
LGTM

dqminh:
LGTM. Delegating the generation of apparmor profile to the consumer looks good !

Issue #639, opened 2015-06-18
memory.swappiness may not be writeable on some kernels


estesp:
PR #622 adding the capability for writing to the `memory.swappiness` setting within a cgroup.  Because of the check (>= 0 and <= 100), even if the cgroup object has no intended setting, by default, I believe it will be 0 (which is a Go default for `int64`, and, if using from Docker, there is no initial setting in docker's "execdriver" template), and therefore, every container start will try and write to the cgroup sysfs entry.

This has two problems: 1) I'm going to assume that the intent of adding the behavior was not to always set swappiness to zero on every container/cgroup (which is I think what happens right now), and 2) setting memory.swappiness in hierarchical cgroups was disabled (you get EINVAL) until this patch: torvalds/linux@3dae7fec5e884a4e72e5416db0894de66f586201 which appears to be in 3.16 and newer kernels.

On Ubuntu 14.04.02 LTS for example, using Docker with an updated libcontainer, even without the "UI" elements for interacting with `memory.swappiness`, every container start fails because of the cgroup write EINVAL result from trying to write zero to `memory.swappiness` on that older kernel.

A fix for problem 1 would be to somehow default an "unset" `MemorySwappiness` to `-1` so writes aren't inadvertently done every time with a zero value.  I think we need to discuss problem 2 as far as either graceful fallback on failure with EINVAL or removing the feature until it is available in the lowest supported kernel.

rjnagal:
-1 for 1) sounds good.
For the second part, I think libcontainer should return an error if a user asks for swappiness when its not available. Docker daemon can ignore requests when swappiness is not enabled, just the way it does for mem swap.

dqminh:
WDYT about changing the default type for cgroups config value to `*int64`. It's a bit more troublesome to consume, but it allows us to properly detect whether values are set by user.

dqminh:
For 2), I'm not sure how we can properly detect this for older kernel. If 1) is fixed, then i think that failing to set user-defined swappiness value is a fatal error and the container should not be started (i.e., there's no approriate fallback)

ktraghavendra:
@estesp @dqminh @cpuguy83 
It is correct that we do not want to set the mem swappiness value to zero for everybody.

Assuming that failing to set the swappiness value when user has asked to do so..
we have
case A: explicit setting of memory swappiness value on a supported kernel
Container does not start
case B: explicit setting of memory swappiness on a  unsupported kernel
Container does not start

[...]

PR #654, opened 2015-06-29
Treat -1 as default value for memory swappiness

ktraghavendra:
As mentioned in #639 for some older kernels setting swappiness
fails. This happens even when nobody tries to configure swappiness
from docker UI because we would still get some default value from
host config.
With this we treat -1 value as default value and skip the enforcement
of swappiness.

Signed-off-by: Raghavendra K T raghavendra.kt@linux.vnet.ibm.com

### DIFF
--- a/cgroups/fs/memory.go
+++ b/cgroups/fs/memory.go
@@ -75,6 +75,10 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 		if err := writeFile(path, "memory.swappiness", strconv.FormatInt(cgroup.MemorySwappiness, 10)); err != nil {
 			return err
 		}
+	} else if cgroup.MemorySwappiness == -1 {
+		return nil
+	} else {
+		return fmt.Errorf("invalid value:%d. valid memory swappiness range is 0-100", cgroup.MemorySwappiness)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


cgroups/fs/memory.go
```diff
@@ -75,6 +75,10 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 		if err := writeFile(path, "memory.swappiness", strconv.FormatInt(cgroup.MemorySwappiness, 10)); err != nil {
 			return err
 		}
+	} else if cgroup.MemorySwappiness == -1 {
+		return nil
+	} else {
+		return fmt.Errorf("Invalid value:%d. Valid memory swappiness range is 0-100", cgroup.MemorySwappiness)
```
vmarmol: nit: Go errors start with lowercase. Here and below.

cgroups/systemd/apply_systemd.go
```diff
@@ -487,6 +487,10 @@ func setKernelMemory(c *configs.Cgroup) error {
 		if err != nil {
 			return err
 		}
+	} else if c.MemorySwappiness == -1 {
```
vmarmol: How is this related to the KernelMemory limit?

LK4D4:
I think you should port this PR to opencontainers/runc/libcontainer.

cgroups/systemd/apply_systemd.go
ktraghavendra: @vmarmol 
Thanks for the review
corrected both error format and the patching problem (KernelMemLimit) that happened while porting from the docker environment to libcontainer

ktraghavendra:
@LK4D4 Thanks, okay will lok into opencontainers' porting

PR #657, opened 2016-11-22
error strings should not be capitalized or end with punctuation

yu8833:
Signed-off-by: yupeng <yu.peng36@zte.com.cn>


hqhq:
Development is moved to https://github.com/opencontainers/runc

yu8833:
@hqhq thanks.

